PR #29, opened 2023-03-13
Add swiftly installation script

### LINKED ISSUE #8: Implement bootstrapping script for Ubuntu variants
For the MVP, we'll need a bootstrapping script that can detect the platform it's being run from (for starters, just Ubuntu flavors) and then download the appropriate `swiftly` executable. This is inspired by [`rustup-init.sh`](https://github.com/rust-lang/rustup/blob/master/rustup-init.sh).

patrickfreed:
Closes #8 

This PR introduces a script that can be used to install swiftly itself. It's intended to be downloaded using curl (or an equivalent) and piped to a shell for quick installation, and its behavior models that of [`rustup-init.sh`](https://github.com/rust-lang/rustup/blob/master/rustup-init.sh) very closely. A subsequent PR will introduce the landing page that provides a conveniently copyable command for this, similar to https://rustup.rs.

This PR also adds a series of bash tests for this script, though I've opted not took hook them up to GitHub actions since I'll be porting them over to Swift CI anyways. I've manually verified that they all pass on the supported Ubuntu flavors and Amazon Linux 2.


install/tests/default-install.sh
```diff
@@ -0,0 +1,35 @@
...
+
+if [[ ! -d "$HOME/.local/share/swiftly/toolchains" ]]; then
+    test_fail "the toolchains directory was not created in SWIFTLY_HOME_DIR"
+fi
+
+swiftly install latest
+
+swift --version
+
+test_pass
```
patrickfreed, line 35: The `test_pass` function used to print some stuff that I ended up moving to `run-tests.sh`, so now it doesn't really do anything. I figured I'd leave it in if we ever wanted to do some passed-test specific stuff though.

install/run-tests.sh
```diff
@@ -0,0 +1,45 @@
+#!/usr/bin/env bash
+
+# Script used to execute all the swiftly-install tests found in install/tests.
+# This should be run from the install directory of the repository.
+# WARNING: these tests make changes to the local filesystem and are intended to be run in a containerized environment.
+
+source ./test-util.sh
```
adam-fowler, line 7: Is it possible to run tests without messing up your current swiftly install?

adam-fowler (commented):
This all looks good.
I verified it worked on my laptop. 
The swiftly exe is much smaller than I expected.
I've added just the one question

adam-fowler:
After this is merged, can you make a release with the shell script attached so people can download it using curl.

install/run-tests.sh
patrickfreed: We could set `SWIFTLY_HOME_DIR` to a temporary directory, but these tests also exercise the default installation path, which inevitably messes with the system install. I think it's worth it to have the tests modify the local system in order to avoid the risk of missing test coverage, especially since they're relatively easy to just run on a fresh docker container.

patrickfreed:
>Is it worthwhile at the end of swiftly install that you ask do you want to install the latest version of swift?

rustup has this behavior by default, and it can be disabled via a flag, but I opted to leave it off for now. The main reason was that Swift toolchain downloads are huge, so if a user doesn't want the latest stable version, it would be best if we don't opt them into installing it. A prompt doesn't mean opting them in, but it would if the `-y` flag was specified, which might be problematic in certain situations. Definitely open to revisiting this decision in the future though. For now, I've added a line of output to the end of installation that demonstrates how a user could install the latest toolchain.

adam-fowler (approved):

adam-fowler:
Is there a plan to include the dependency installation into this script, or is that going to go somewhere else?

adam-fowler:
> Is there a plan to include the dependency installation into this script, or is that going to go somewhere else?

I added #32 to track this

PR #40, opened 2023-04-03
Capitalized Swiftly in README

MahdiBM:
I personally don't have strong opinions about these kinds of things as long as the text is understandable, but Swiftly as a name should be capitalized as far as I know.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
-# swiftly
+# Swiftly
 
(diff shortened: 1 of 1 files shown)


ktoso:
That's probably a good idea 👍

patrickfreed:
So I realize this is a bit of a personal preference thing, but the all lowercase was actually done on purpose, see [my comment](https://github.com/swift-server/swiftly/pull/12#discussion_r984473813) here for a bit of my reasoning. Given that you're not the first person to bring this up, I wonder if it would be better to refer to it either as `swiftly` then (like `swift-format`: https://github.com/apple/swift-format, notably not like SwiftFormat though), since I certainly don't want to give the impression the README was written lazily or anything.

ktoso:
Hm, since on purpose then maybe a small note somewhere about it in the readme to point people at :-)

MahdiBM:
I like "Swiftly" more since that's more prestigious 😅, but I see where you're coming from.
I think we can use backticks (e.g. `swiftly`) like swift-format does, and that should be enough proof for readers that this is intentional.

Issue #48, opened 2023-04-07
If installing as root use different PATH from ~/.local/bin


adam-fowler:
In general root users do not have `~/.local/bin` in their PATH. If the user is installing as root perhaps we should use a different PATH. I think we should also warn the user when installing as root.

patrickfreed:
>In general root users do not have ~/.local/bin in their PATH. If the user is installing as root perhaps we should use a different PATH.

I think always following the [XDG base directory specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html) is still the right approach even for the root user, since installing something as root doesn't necessarily imply that you want that installation available to all users on a system. It is true that root doesn't have `$HOME/.local/bin` in their PATH by default, but #42 should make that situation not really a big deal anyways. Another important thing to consider is that the Swift toolchain bundles a bunch of executables with it that are more likely to clash if installed to a directory like `/usr/local/bin`, and the consequences of clashing in a system-wide directory like that are a bit more serious too.

> I investigated what rustup does, and what I found is:
> 
> - First you curl | sh the rustup installer.
> - The installer downloads another installer considering your platform.
> - That other installer [has some code](https://github.com/rust-lang/rustup/blob/master/src/cli/self_update/shell.rs) to find proper rc files.
> - They then set up the rc files to run their env script on every login (. "$HOME/.cargo/env").
> 
> The problem is setting up swiftly's installation script to do all these is a little bit weird/hard to me. Definitely doable though.

Yeah, this is something @adam-fowler suggested a while back. I've held off on doing it so far since I didn't want to accidentally clobber a user's shell configuration, but I think it's something we'll want to do eventually. I'll investigate it as part of #42 to see how much work is involved there. If it's not too bad, we can do it as part of that and otherwise split it off and address it in the future.

>We might be able to do something like this:

This is an interesting idea, but I think it's best if we stick to a simpler approach that follows the base directory specification. It's a widely known and familiar spec that makes it so handing things like this is relatively straightforward and standardized. Rust notably doesn't follow the spec right now, but I don't think their example is necessarily one we want to follow in this case (https://github.com/rust-lang/cargo/issues/1734 has 200 thumbs up). I also think that changing where the installation goes depending on the value of PATH might be surprising to some users.

So all that being said, I'm inclined to say that it's probably best that we don't do this, but that once we do #42, this issue will be less of a concern.

patrickfreed:
Now that #42 has been completed, I think we can close this.

PR #49, opened 2023-04-07
`swiftly` installer change default bin to `/usr/local/bin`

MahdiBM:
Using the changes in this PR, swiftly and its installer will default to `/usr/local/bin` instead of `$HOME/.local/bin`.

I was initially going to do [this](https://github.com/swift-server/swiftly/issues/48#issuecomment-1500279519), but ended up doing this after some more research.

I personally checked 6 different OS instances, (my MacBook + Ubuntu 22.04 & 18.04 & amazonlinux on Docker + Ubuntu 22.04 on 2 vps, one with root and one with a default user). None of them actually had `$HOME/.local/bin` in their PATH and all of them had `/usr/local/bin` (Not even the 2 non-root-user OS: my MacBook and a vps).

I did find this [Github Issue](https://github.com/JuliaLang/juliaup/issues/247) that mentions there should be a `$HOME/.local/bin`, but I just don't see it. Overall i'm a little bit confused of the PATH situation, but I think `/usr/local/bin` should solve the problem anyway.

The [bash docs](https://git.savannah.gnu.org/cgit/bash.git/tree/doc/bash.1?h=f188aa6a013e89d421e39354086eed513652b492#n2501) says this about PATH:
> .B PATH
indicates the current directory.
A null directory name may appear as two adjacent colons, or as an initial
or trailing colon.
The default path is system-dependent,
and is set by the administrator who installs
.BR bash .
A common value is
.na
.if t \f(CW/usr/local/bin:/usr/local/sbin:/usr/bin:/usr/sbin:/bin:/sbin\fP.
.if n ``/usr/local/bin:/usr/local/sbin:/usr/bin:/usr/sbin:/bin:/sbin''.

This, plus my own observations, makes me think we can put some trust in `/usr/local/bin`.

Solves #48.

### DIFF
--- a/Sources/SwiftlyCore/Platform.swift
+++ b/Sources/SwiftlyCore/Platform.swift
@@ -67,13 +67,11 @@ extension Platform {
     ///
     /// If a mocked home directory is set, this will be the "bin" subdirectory of the home directory.
     /// If not, this will be the SWIFTLY_BIN_DIR environment variable if set. If that's also unset,
-    /// this will default to ~/.local/bin.
+    /// this will default to /usr/local/bin.
     public var swiftlyBinDir: URL {
-        SwiftlyCore.mockedHomeDir.map { $0.appendingPathComponent("bin", isDirectory: true) }
-            ?? ProcessInfo.processInfo.environment["SWIFTLY_BIN_DIR"].map { URL(fileURLWithPath: $0) }
-            ?? FileManager.default.homeDirectoryForCurrentUser
-            .appendingPathComponent(".local", isDirectory: true)
-            .appendingPathComponent("bin", isDirectory: true)
+        SwiftlyCore.mockedHomeDir.map { $0.appendingPathComponent("bin", isDirectory: true) } ??
+          ProcessInfo.processInfo.environment["SWIFTLY_BIN_DIR"].map { URL(fileURLWithPath: $0) } ?? 
+          URL(fileURLWithPath: "/usr/local/bin")
     }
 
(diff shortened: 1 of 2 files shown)


patrickfreed:
Thanks for making this PR, but unfortunately I don't necessarily think this is the direction we want to go. See my comment on #48 for a bit more context.

Regarding `~/.local/bin` not being on the PATH, that is a bit surprising, since it is used by lots of different things for storing executables due to its definition in the base directory specification. Ubuntu 20.04 at least adds it to the PATH by default if it exists in the `.profile` file shipped with it:
```bash
# ~/.profile: executed by the command interpreter for login shells.
# This file is not read by bash(1), if ~/.bash_profile or ~/.bash_login
# exists.
# see /usr/share/doc/bash/examples/startup-files for examples.
# the files are located in the bash-doc package.

# the default umask is set in /etc/profile; for setting the umask
# for ssh logins, install and configure the libpam-umask package.
#umask 022

# if running bash
if [ -n "$BASH_VERSION" ]; then
    # include .bashrc if it exists
    if [ -f "$HOME/.bashrc" ]; then
	. "$HOME/.bashrc"
    fi
fi

# set PATH so it includes user's private bin if it exists
if [ -d "$HOME/bin" ] ; then
    PATH="$HOME/bin:$PATH"
fi

# set PATH so it includes user's private bin if it exists
if [ -d "$HOME/.local/bin" ] ; then
    PATH="$HOME/.local/bin:$PATH"
fi
```

PR #51, opened 2023-04-12
Fix interactive input for `curl ... | bash` style installation

kateinoigakukun:
In that installation style, `read_input_with_default` returned always the default value without waiting for user input because the stdin is piped from curl and it's already EOF. This patch fixes it by reading `/dev/tty` as stdin explicitly.

### DIFF
--- a/install/swiftly-install.sh
+++ b/install/swiftly-install.sh
@@ -28,7 +28,8 @@ read_input_with_default () {
     # The installer script is usually run by "curl ... | bash", which means that
     # stdin is not a tty but the script content itself. In that case, "read" builtin
     # command receives EOF immediately. To avoid that, we use /dev/tty as stdin explicitly.
-    if [ -t 0 ]; then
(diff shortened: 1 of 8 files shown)


patrickfreed:
@swift-server-bot add to allowlist

patrickfreed:

patrickfreed:
hmm, @yim-lee it looks I might not have permission to add users to the allow list.

patrickfreed:
@swift-server-bot add to allowlist

patrickfreed:

patrickfreed:
looks like it worked, thanks again!

patrickfreed (approved):
LGTM, thanks for catching this and fixing it!

patrickfreed:
Actually, I just noticed a minor error is printed when `SWIFTLY_READ_FROM_STDIN` isn't set, so I went ahead and fixed that. I also updated `test-util.sh` to set that variable rather than doing in the tests themselves--that way we won't have to remember to set it when we read user input.

patrickfreed:

PR #53, opened 2023-04-25
Provide an `env.sh` file that can be sourced, automatically update shell profile config file to source it

### LINKED ISSUE #42: Provide an `env` file that can be used to quickly set up swiftly installation
rustup provides an `env` file which can be used to easily set environment variables used by rustup, including amending the PATH. We should consider providing a similar file for swiftly installations, since right now we just print a message telling the user they should do it themselves.

patrickfreed:
Closes #42 

This PR updates `swiftly-install.sh` to create `$SWIFTLY_HOME_DIR/env.sh`, which contains the required environment variable definitions needed to run swiftly. After installation, users can source it to enable using swiftly in their current shell easily.

This PR also updates `swiftly-install.sh` to optionally modify the user's login shell configuration profile (on by default) to include a line that sources the aforementioned `env.sh` file. For right now, only one of `.profile`, `.bash_profile`, `.bash_login`, or `.zprofile` will be modified, depending on the value of `$SHELL` and the existence of said files. In other words, only login configuration files will be modified but not any per-shell rc files like `.bashrc`. The reasons for this are twofold: 1) environment variables that are inherited by subshells are supposed to be defined in a login configuration file, not an rc file, since otherwise they would get redefined in each subshell, possibly recursively (as is the case with $PATH), 2) this approach is more conservative and can be extended in the future should `.profile` and friends prove to be insufficient. rustup takes the opposite approach of putting `. `$HOME/.cargo/env` in every profile or rc file it can find, but that seems like it might be overkill.


patrickfreed:

patrickfreed:

adam-fowler (approved):
Looks good to me

patrickfreed:

gregcotten:
This has never worked for me on any clean vanilla Ubuntu 20, 22, or 24 installs. I'm not entirely sure why, but the "Terminal" app in Ubuntu simply doesn't load the .profile file by default. So I run the swiftly curl install incantation and have to manually add the env.sh invocation in the user's .bashrc

gregcotten:
> rustup takes the opposite approach of putting . $HOME/.cargo/env` in every profile or rc file it can find, but that seems like it might be overkill.

There might be a reason for this :)

PR #54, opened 2023-05-28
Install system dependencies in swiflty-install.sh

### LINKED ISSUE #32: Linux: Download required dependencies
See https://www.swift.org/download/#linux

patrickfreed:
Closes #32

This works by scraping the dependency list from the dockerfiles at https://github.com/apple/swift-docker. Long-term, we'll want to get the list of dependencies from a swift.org API.

The design originally called for us to not do this due to different Swift releases having different system dependencies, but upon further reflection, this seems like a big enough ergonomic win that we might as well do it.

### DIFF
--- a/install/swiftly-install.sh
+++ b/install/swiftly-install.sh
@@ -22,6 +22,9 @@
 # Unless the --disable-confirmation flag is set, this script will allow the runner to
 # configure either of those two directory paths.
 #
+# Unless the --no-install-system-deps flag is set, this script will attempt to install Swift's
+# system dependencies using the system package manager.
+#
 # curl is required to run this script.
 
 set -o errexit
@@ -59,25 +62,34 @@ yn_prompt () {
 # Read a y/n input.
 # First argument is the default value (must be "true" or "false").
 #
-# Sets READ_INPUT_RETURN to "true" for an input of "y" or "Y" and sets it to
-# "false" for an input of "n" or "N". For all other inputs, READ_INPUT_RETURN is
-# set to the default value provided in the first argument.
+# Sets READ_INPUT_RETURN to "true" for an input of "y" or "Y", "false" for an input
+# of "n" or "N", or the default value for a blank input
+# 
+# For all other inputs, a message is printed and the user is prompted again.
 read_yn_input () {
-    read_input_with_default "$1"
+    while [[ true ]]; do
+        read_input_with_default "$1"
 
-    case "$READ_INPUT_RETURN" in
-        "y" | "Y")
-            READ_INPUT_RETURN="true"
-            ;;
+        case "$READ_INPUT_RETURN" in
+            "y" | "Y")
+                READ_INPUT_RETURN="true"
+                return
+                ;;
 
-        "n" | "N")
-            READ_INPUT_RETURN="false"
-            ;;
+            "n" | "N")
+                READ_INPUT_RETURN="false"
+                return
+                ;;
 
-        *)
-            READ_INPUT_RETURN="$1"
-            ;;
-    esac
+            "$1")
+                return
+                ;;
+
+            *)
+                echo "Please input either \"y\" or \"n\", or press ENTER to use the default."
+                ;;
+        esac
+    done
 }
 
 # Replaces the actual path to $HOME at the beginning of the provided string argument with
@@ -120,7 +132,7 @@ install_system_deps () {
     dockerfile_url="https://raw.githubusercontent.com/apple/swift-docker/main/nightly-main/$docker_platform_name/$docker_platform_version/Dockerfile"
     dockerfile="$(curl --silent --retry 3 --location --fail $dockerfile_url)"
     if [[ "$?" -ne 0 ]]; then
-        echo "Error enumerating system dependencies, skipping system dependencies installation."
+        echo "Error enumerating system dependencies, skipping installation of system dependencies."
     fi
 
     # Find the line number of the RUN command associated with installing system dependencies.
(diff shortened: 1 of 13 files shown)


patrickfreed:

patrickfreed:

patrickfreed:

install/swiftly-install.sh
```diff
@@ -48,6 +51,47 @@ read_input_with_default () {
     fi
 }
 
+yn_prompt () {
```
patrickfreed, line 54: I originally thought we needed another y/n prompt, so factored it out into these functions. We didn't end up needing it, but it still seemed useful so I left it in.

install/swiftly-install.sh
```diff
@@ -71,9 +115,57 @@ bold () {
...
+install_system_deps () {
+    if [[ "$(id --user)" != "0" ]] && ! has_command sudo ; then
+        echo "Warning: sudo not installed and current user is not root, skipping system dependency installation."
+        return
+    elif ! has_command "$package_manager" ; then
+        echo "Warning: package manager \"$package_manager\" not found, skipping system dependency instllation."
+        return
+    fi
+
+    dockerfile_url="https://raw.githubusercontent.com/apple/swift-docker/main/nightly-main/$docker_platform_name/$docker_platform_version/Dockerfile"
```
patrickfreed, line 132: The system dependencies can change on each release, so I decided to just use the latest set of them to ensure a user can always install the latest swift version successfully. I thought about unioning the oldest releases' dependencies with the newest one's, but that seemed overly complex for an uncommon case.

install/swiftly-install.sh
```diff
@@ -71,9 +115,57 @@ bold () {
...
+    dockerfile="$(curl --silent --retry 3 --location --fail $dockerfile_url)"
+    if [[ "$?" -ne 0 ]]; then
+        echo "Error enumerating system dependencies, skipping installation of system dependencies."
+    fi
+
+    # Find the line number of the RUN command associated with installing system dependencies.
+    beg_line_num=$(printf "$dockerfile" | grep -n --max-count=1 "$package_manager.*install" | cut -d ":" -f1)
+
+    # Starting from there, find the first line that doesn't have the same level of indentation.
+    relative_end_line_num=$(printf "$dockerfile" |
```
patrickfreed, line 142: I started using lowercase variable names for non-environment variables per https://stackoverflow.com/a/673940/16745421. In a later commit, I'll go back and update the existing variables to match this.

install/tests/disable-prompt.sh
```diff
@@ -15,7 +15,13 @@ cleanup () {
 }
 trap cleanup EXIT
 
-./swiftly-install.sh -y
+if has_command apt-get ; then
+    apt-get remove -y zlib1g-dev
+elif has_command yum ; then
+    yum remove -y libcurl-devel
```
patrickfreed, line 21: These are just arbitrary packages from the dependency lists used to check if the whole set have been installed or not.

install/swiftly-install.sh
```diff
@@ -71,9 +115,57 @@ bold () {
...
+# repository and attempt to install them using the system's package manager.
+#
+# $docker_platform_name, $docker_platform_version, and $package manager need
+# to be set before calling this function.
+install_system_deps () {
+    if [[ "$(id --user)" != "0" ]] && ! has_command sudo ; then
+        echo "Warning: sudo not installed and current user is not root, skipping system dependency installation."
+        return
+    elif ! has_command "$package_manager" ; then
+        echo "Warning: package manager \"$package_manager\" not found, skipping system dependency instllation."
```
adam-fowler, line 128:
```suggestion
        echo "Warning: package manager \"$package_manager\" not found, skipping system dependency installation."
```

install/swiftly-install.sh
```diff
@@ -71,9 +115,57 @@ bold () {
...
+    dockerfile_url="https://raw.githubusercontent.com/apple/swift-docker/main/nightly-main/$docker_platform_name/$docker_platform_version/Dockerfile"
+    dockerfile="$(curl --silent --retry 3 --location --fail $dockerfile_url)"
+    if [[ "$?" -ne 0 ]]; then
+        echo "Error enumerating system dependencies, skipping installation of system dependencies."
+    fi
+
+    # Find the line number of the RUN command associated with installing system dependencies.
+    beg_line_num=$(printf "$dockerfile" | grep -n --max-count=1 "$package_manager.*install" | cut -d ":" -f1)
+
+    # Starting from there, find the first line that doesn't have the same level of indentation.
```
adam-fowler, line 141: Indentation seems a little flaky way to catch the full list of installation dependencies. Why don't you use the fact that the last non-whitespace character is a backslash. Maybe once you have found the install line concatenate all the lines ending with `\` together. Then split the line into separate commands ie with `&&` as separator. Then choose the install command and read the packages from it.

Its still not ideal and just hardcoding the required packages might just be preferable.

If you are going to parse the Docker file then I'd also add tests for the parsing.

install/swiftly-install.sh
```diff
@@ -71,9 +115,57 @@ bold () {
...
+
+    install_args=(--quiet)
+    if [[ "$DISABLE_CONFIRMATION" == "true" ]]; then
+        install_args+=(-y)
+    fi
+
+    # Disable errexit since failing to install system dependencies is not swiftly installation-fatal.
+    set +o errexit
+    if [[ "$(id --user)" == "0" ]]; then
+        "$package_manager" install "${install_args[@]}" "${package_list[@]}"
```
adam-fowler, line 158: Is there a chance on ubuntu that you need to call `apt-get update` first?

install/swiftly-install.sh
patrickfreed: fixed

install/swiftly-install.sh
patrickfreed: Yeah, indentation might not be be the best thing to rely on. Updated to do something more along the lines of your suggestion. I don't really want to hardcode any lists of dependencies since they vary from platform to platform and are updated every now and then.

The tests (in particular default-install.sh) do cover the parsing behavior, though not as comprehensively as they should. I've updated default-install.sh to verify more of the required dependencies were installed. For that I just hardcoded the lists, which is okay since they're just tests. We may need to update it from time to time, but the cost of doing so shouldn't be too bad.

install/swiftly-install.sh
patrickfreed: Yeah, especially for a fresh docker image. That said, I think it would be better for users to do that outside of the script, since in most cases calling update shouldn't be needed, and we don't want to do it unnecessarily since it can involve downloading a decent amount of data.

patrickfreed:

FranzBusch:
@etcwilde @MaxDesiatov  This might be of interested to our recent discussion in https://github.com/apple/swift/pull/66398 as well.

install/swiftly-install.sh
adam-fowler: If a package install fails is it worthwhile pointing out to the user they might want to run `apt-get update`

install/swiftly-install.sh
adam-fowler: My regex, linux commands knowledge is a little rusty so I'm gonna have to trust you on this. But much prefer checking for `\` than the tabbing.

adam-fowler (approved):

patrickfreed:

install/swiftly-install.sh
patrickfreed: good idea, done

patrickfreed:

PR #56, opened 2023-08-21
Always disable confirmation when installing system dependencies

patrickfreed:
This fixes a bug similar to the one identified in #51, wherein `curl ... | bash` installation would fail due to stdin being EOF. In this case, `apt-get` was hanging because it needed some input to install the `tzdata` package. To address this, I've updated the system deps installation phase to just always proceed without confirmation. If a user wants to disable the installation of system deps, they already had to confirm as much during the options prompting phase.


patrickfreed:

PR #57, opened 2023-08-22
Website for Swiftly

adam-fowler:
Install page for Swiftly


patrickfreed:
Thanks for the PR, sorry I never got around to doing this. I made a few minor edits to it, the most noticeable of which was making the logo smaller. I figured we want the main focus to be on the one liner, so it making the logo smaller helps with that. Let me know what you think.

adam-fowler:
I resized the actual swift png. No point loading 400x400 and rendering at a much smaller size. I think we should have it at 100x100. 50x50 is too small

patrickfreed:
ok cool, thank you! looks great to me

PR #58, opened 2023-08-25
Accept SNAPSHOT in toolchain selector parsing

patrickfreed:
This PR changes `ToolchainSelector` parsing to now accept inputs such as `5.7-SNAPSHOT` or `main-SNAPSHOT`. I opted not to add `swift-SNAPSHOT` as an alternative for `main-snapshot` since it would be another set of words we'd be accepting, whereas `main-SNAPSHOT` is only a difference of capitalization from the existing `main-snapshot`.


ktoso:
Hooray thank you!

adam-fowler (approved):

Issue #21, opened 2022-10-30
Implement macOS support


patrickfreed:
This is a tracking issue for macOS support. More details will be added as that work is scoped out.

Once Linux has reached MVP status, this will be addressed.

iainsmith:
HI @patrickfreed, I was looking at this quickly and was wondering what's needed for this:

- Update Package.swift
- Implement MacPlatform.swift and `struct Mac: Platform` (or would you prefer Darwin) as the name here?

I'd be keen to look at an existing branch, if one exists.

sdavids:
https://github.com/swiftlang/swiftly/blob/364b02bb2904cff62a604bfa4a0ef0d1d9164cd5/DESIGN.md?plain=1#L57


Please honor XDG conventions on MacOS also—Mac users prefer uncluttered home directories too 😉

cmcgee1024:
This has been implemented in #121

PR #62, opened 2023-09-20
[Feature] Add initial macOS support for compiling

Kyle-Ye:
## Related issue:

#21

## Description

This PR add initial Darwin platform support for compiling.

Many feature is missing and it does not mean we support macOS on this PR.

## TODO & Future Plan

- [ ] Add `name` getter in `Platform` protocol as described in Design.md
- [ ] Extract the same logic in `Linux` and `Darwin` to SwiftlyCore via `extension Platform`
- [ ] Update Darwin implementation to better respect the OS - toolchain install path / file extension / Xcode integration

### DIFF
--- a/Sources/DarwinPlatform/Darwin.swift
+++ b/Sources/DarwinPlatform/Darwin.swift
@@ -2,10 +2,10 @@ import Foundation
 import SwiftlyCore
 @_implementationOnly import Archive
 
-/// `Platform` implementation for Linux systems.
-/// This implementation can be reused for any supported Linux platform.
+/// `Platform` implementation for Darwin systems.
+/// This implementation can be reused for any supported Darwin platform.
 /// TODO: replace dummy implementations
-public struct Linux: Platform {
+public struct Darwin: Platform {
     public init() {}
 
     public var appDataDirectory: URL {
@@ -19,7 +19,7 @@ public struct Linux: Platform {
     }
 
     public var toolchainFileExtension: String {
-        "tar.gz"
+        "pkg"
     }
 
     public func isSystemDependencyPresent(_: SystemDependency) -> Bool {
@@ -140,5 +140,5 @@ public struct Linux: Platform {
         FileManager.default.temporaryDirectory.appendingPathComponent("swiftly-\(UUID())")
     }
 
-    public static let currentPlatform: any Platform = Linux()
+    public static let currentPlatform: any Platform = Darwin()
 }
--- a/Package.swift
+++ b/Package.swift
@@ -4,6 +4,8 @@ import PackageDescription
 
 let package = Package(
     name: "swiftly",
+    // Current supported Darwin family: macOS
+    platforms: [.macOS(.v13)],
     products: [
         .executable(
             name: "swiftly",
@@ -23,6 +25,7 @@ let package = Package(
                 .product(name: "ArgumentParser", package: "swift-argument-parser"),
                 .target(name: "SwiftlyCore"),
                 .target(name: "LinuxPlatform", condition: .when(platforms: [.linux])),
+                .target(name: "DarwinPlatform", condition: .when(platforms: [.macOS])),
                 .product(name: "SwiftToolsSupport-auto", package: "swift-tools-support-core"),
             ]
         ),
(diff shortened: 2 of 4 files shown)


Sources/DarwinPlatform/Darwin.swift
```diff
@@ -0,0 +1,144 @@
+import Foundation
+import SwiftlyCore
+@_implementationOnly import Archive
```
adam-fowler, line 3: Why is this @_implementationOnly? I'd rather we didn't include `_` directives if they can be avoided.

Sources/DarwinPlatform/Darwin.swift
```diff
@@ -0,0 +1,144 @@
...
+        guard tmpFile.fileExists() else {
+            throw Error(message: "\(tmpFile) doesn't exist")
+        }
+
+        if !self.swiftlyToolchainsDir.fileExists() {
+            try FileManager.default.createDirectory(at: self.swiftlyToolchainsDir, withIntermediateDirectories: false)
+        }
+
+        SwiftlyCore.print("Extracting toolchain...")
+        let toolchainDir = self.swiftlyToolchainsDir.appendingPathComponent(version.name)
```
adam-fowler, line 39: Should toolchain directory be `~/Library/Developer/Toolchains` as this is the standard local directory to use for Toolchain installs with Xcode

Sources/DarwinPlatform/Darwin.swift
```diff
@@ -0,0 +1,144 @@
...
+
+        if toolchainDir.fileExists() {
+            try FileManager.default.removeItem(at: toolchainDir)
+        }
+
+        try extractArchive(atPath: tmpFile) { name in
```
adam-fowler, line 45: Does this work on `pkg` files

Sources/LinuxPlatform/Linux.swift
```diff
@@ -1,5 +1,6 @@
 import Foundation
 import SwiftlyCore
+@_implementationOnly import Archive
```
adam-fowler, line 3: Again why @_implementationOnly.

adam-fowler (commented):
This is only a very quick review from an initial look at the code.

Sources/DarwinPlatform/Darwin.swift
Kyle-Ye: I think Archive(Or CLibArchive) is an internal detail of the implementation. So I use `@_implementationOnly import` to imply the fact and also prevent use from using CLibArchive symbol in any public API of the module in compile time.

`@_implementation import` and `@_exported import` is widely used in Swift ecosystem and there is some discussion in the forum to make them stable - remove "_".

But since swiftly currently only provide an executable produce and would not be another package's dependency. It's OK to use the normal import.

Sources/DarwinPlatform/Darwin.swift
Kyle-Ye: Yes. We should have done it. The implementation of `DarwinPlatform` is mostly duplicated from `LinuxPlatform`. I'd like to leave it to a future PR/improvement.(This PR only promise we can build it on macOS platform instead of making it usable on macOS)
But if you think the current code status does not meet the merge standard, I can also address it in this PR.

Sources/DarwinPlatform/Darwin.swift
Kyle-Ye: Not sure.
The same reason as last one. https://github.com/swift-server/swiftly/pull/62#discussion_r1331905196

Sources/LinuxPlatform/Linux.swift
Kyle-Ye: See https://github.com/swift-server/swiftly/pull/62#discussion_r1331905196

patrickfreed:
Hi @Kyle-Ye, thanks for taking the time to submit this PR! The macOS implementation is likely to look a bit different than the Linux one though, so I think we'll want to hold off on duplicating too much code between the platforms. Also, I think we'll want to wait on advertising macOS support, even in just Package.swift, until we begin the actual implementation, since having a placeholder risks a bit of confusion if people don't know the status of the project currently. So given that, I'm going to close this PR for now, since I think we'll want to do this at the same time as we begin implementing the actual logic.

Issue #60, opened 2023-09-18
Allow installation on derived and unsupported platforms


patrickfreed:
To enable users to use swiftly on platforms that aren't explicitly supported by Swift, we could provide the option during swiftly installation to use the toolchains intended for another supported platform. 

e.g. If installing on debian, a user could select Ubuntu 20.04 as the platform they'd like their toolchains to download from, and it's likely they would work ok (I haven't tried this though).

makoni:
Another case - Ubuntu 23.04. I can't install Swift using swiftly, had to do a manual installation of Swift for 22.04 :)

Diggory:
Yes please. Anything that makes it easy to install and update Swift on Raspberry Pi OS would be very gratefully received.

adam-fowler:
I think this can be split into two tasks. 

1) The user explicitly indicates the platform they want to install via a command line parameter 
2) reading the ID_LIKE variable from os-release file.

patrickfreed:
I think we may also want to introduce a CLI option for the platform too, which will skip the entire platform detection process. This will help when installing swiftly as part of a CI task, for example.

I'll try to get a PR up for this tonight.

patrickfreed:
This has been implemented and deployed!

PR #63, opened 2023-09-21
Support installation on alternative platforms

patrickfreed:
Implements #60

### DIFF
--- a/install/swiftly-install.sh
+++ b/install/swiftly-install.sh
@@ -170,7 +170,7 @@ install_system_deps () {
     set -o errexit
 }
 
-set_ubuntu_platform () {
+set_platform_ubuntu () {
     docker_platform_name="ubuntu"
     package_manager="apt-get"
     export DEBIAN_FRONTEND=noninteractive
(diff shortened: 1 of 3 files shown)


patrickfreed:

patrickfreed:

install/swiftly-install.sh
```diff
@@ -170,26 +167,198 @@ install_system_deps () {
...
+            ;;
+
+        *)
+            echo "Cancelling installation."
+            exit 0
+            ;;
+    esac
+}
+
+verify_getopt_install () {
```
patrickfreed, line 308: `getopt` is a super useful tool for parsing CLI options in bash scripts. It's included in the util-linux package, but in practice basically every Linux distro comes with it pre-installed...except Amazon Linux 2. The alternatives are manual parsing (which gets tricky with stuff like `-pubuntu20.04`, `-p  ubuntu20.04`, `--platform=ubuntu20.04`, and `--platform ubuntu20.04`) or using the shell builtin `getopts`, which doesn't support long names like `--platform`. Given that we already require `curl` to be installed, it didn't seem like a huge problem to add an additional dependency on `util-linux`.

install/swiftly-install.sh
```diff
@@ -170,26 +167,198 @@ install_system_deps () {
...
+        return 1
+    fi
+
+    getopt --test
+    # getopt --test exiting with status code 4 implies GNU getopt is being used, which we need.
+    [[ "$?" -eq 4 ]]
+    return "$?"
+}
+
+SWIFTLY_INSTALL_VERSION="0.2.0"
```
patrickfreed, line 319: the swiftly-install version doesn't need to be tied to swiftly's actual version imo, since we may make updates to it like this independent of changes to swiftly.

adam-fowler: We need to keep the tags on the repo in sync though. A new version of the script should be tagged a release, otherwise managing versions of the script is going to be painful.

adam-fowler (approved):
All looks good.

install/swiftly-install.sh
patrickfreed: Yeah, I agree we ought to tag releases of swiftly-install, though we shouldn't mark them as a full on "Releases" in the GitHub sense of things, since then we'd need the install script to disambiguate when trying to download binaries.

PR #67, opened 2023-09-26
Implement update command

### LINKED ISSUE #5: Implement `update` command on Linux
The `update` command accepts a selector and updates the newest matching toolchain to the newest available matching toolchain, if any. This happens in a few steps:

- List installed toolchains to see if any match the selector
  - If no selector is provided, match the currently in-use toolchain
- If one is found, list available toolchains, and see if any matches the selector and is _newer_ than the found toolchain
- Uninstall the found toolchain
- Install the newer toolchain
  - If the updated toolchain was in use, mark the new one as in use

Updating depends on the functionality from a bunch of other commands, so they will need to be done first (#2, #4).

patrickfreed:
Closes #5 

See the README or the `--help` output in `Update.swift` for information on the semantics of the command.

### DIFF
--- a/Sources/Swiftly/Update.swift
+++ b/Sources/Swiftly/Update.swift
@@ -40,34 +40,59 @@ struct Update: SwiftlyCommand {
     ))
     var toolchain: String?
 
+    @Flag(
+        name: [.long, .customShort("y")],
+        help: "Update the selected toolchains without prompting for confirmation."
+    )
+    var assumeYes: Bool = false
+
     public mutating func run() async throws {
-        guard let oldToolchain = try self.oldToolchain() else {
+        var config = try Config.load()
+
+        guard let oldToolchain = try self.oldToolchain(config) else {
             if let toolchain = self.toolchain {
-                print("No installed toolchain matched \"\(toolchain)\"")
+                SwiftlyCore.print("No installed toolchain matched \"\(toolchain)\"")
             } else {
-                print("No toolchains are currently installed")
+                SwiftlyCore.print("No toolchains are currently installed")
             }
             return
         }
 
-        guard let newToolchain = try await self.newToolchain(old: oldToolchain) else {
-            print("\(oldToolchain) is already up to date!")
+        guard let newToolchain = try await self.fetchNewToolchain(old: oldToolchain) else {
+            SwiftlyCore.print("\(oldToolchain) is already up to date")
             return
         }
 
-        print("updating \(oldToolchain) -> \(newToolchain)")
-        try await Install.execute(version: newToolchain)
-        try Swiftly.currentPlatform.uninstall(oldToolchain)
-        print("successfully updated \(oldToolchain) -> \(newToolchain)")
+        guard !config.installedToolchains.contains(newToolchain) else {
+            SwiftlyCore.print("The newest version of \(oldToolchain) (\(newToolchain)) is already installed")
+            return
+        }
+
+        if !self.assumeYes {
+            SwiftlyCore.print("Update \(oldToolchain) ⟶ \(newToolchain)?")
+            guard SwiftlyCore.promptForConfirmation(defaultBehavior: true) else {
+                SwiftlyCore.print("Aborting")
+                return
+            }
+        }
+
+        try await Install.execute(version: newToolchain, &config)
+
+        if config.inUse == oldToolchain {
+            try await Use.execute(newToolchain, &config)
+        }
+
+        try await Uninstall.execute(oldToolchain, &config)
+        SwiftlyCore.print("Successfully updated \(oldToolchain) ⟶ \(newToolchain)")
     }
 
-    private func oldToolchain() throws -> ToolchainVersion? {
+    private func oldToolchain(_ config: Config) throws -> ToolchainVersion? {
         guard let input = self.toolchain else {
-            return try Config.load().inUse
+            return config.inUse
         }
 
         let selector = try ToolchainSelector(parsing: input)
-        let toolchains = try Config.load().listInstalledToolchains(selector: selector)
+        let toolchains = config.listInstalledToolchains(selector: selector)
 
         // When multiple toolchains are matched, update the latest one.
         // This is for situations such as `swiftly update 5.5` when both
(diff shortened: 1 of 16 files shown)


Sources/SwiftlyCore/HTTPClient.swift
```diff
@@ -5,23 +5,45 @@ import NIO
...
+        _ toolchain: ToolchainVersion,
+        url: String,
+        to destination: String,
+        reportProgress: @escaping (SwiftlyHTTPClient.DownloadProgress) -> Void
+    ) async throws
+}
+
 /// HTTPClient wrapper used for interfacing with various APIs and downloading things.
-public class HTTP {
+public class SwiftlyHTTPClient {
```
patrickfreed, line 18: Most of swiftly's test's runtime is devoted to downloading toolchains, and so in an effort to reduce that, I introduced the ability to mock toolchain downloads. This allows the update tests to run quickly while still verifying that updating works. We can consider expanding the use of mocking to other tests to reduce how long they run for, though I do think there is value in having the install command tests be non-mocked.

Sources/SwiftlyCore/HTTPClient.swift
```diff
@@ -5,23 +5,45 @@ import NIO
...
     private static let client = HTTPClientWrapper()
```
adam-fowler, line 19: You create a new HTTPClient with every instance of SwiftlyHTTPClient. Perhaps HTTPClientWrapper should be a wrapper for a global instance.

Sources/Swiftly/Use.swift
```diff
@@ -47,11 +47,11 @@ internal struct Use: SwiftlyCommand {
             return
         }
 
-        try await Self.execute(toolchain, config: &config)
+        try await Self.execute(toolchain, &config)
     }
 
     /// Use a toolchain. This method modifies and saves the input config.
-    internal static func execute(_ toolchain: ToolchainVersion, config: inout Config) async throws {
+    internal static func execute(_ toolchain: ToolchainVersion, _ config: inout Config) async throws {
```
adam-fowler, line 54: Any reason you removed the config label from all the execute functions?

Sources/Swiftly/Update.swift
```diff
@@ -31,61 +38,154 @@ struct Update: SwiftlyCommand {
...
 
+    @Flag(
+        name: [.long, .customShort("y")],
+        help: "Update the selected toolchains without prompting for confirmation."
+    )
+    var assumeYes: Bool = false
+
+    public var httpClient = SwiftlyHTTPClient()
+
+    private enum CodingKeys: String, CodingKey {
```
adam-fowler, line 69: What are the coding keys for?

Sources/Swiftly/Update.swift
```diff
@@ -31,61 +38,154 @@ struct Update: SwiftlyCommand {
...
     }
 
-    private func newToolchain(old: ToolchainVersion) async throws -> ToolchainVersion? {
-        switch old {
-        case let .stable(oldRelease):
-            return try await HTTP.getReleaseToolchains(limit: 1) { release in
-                release.major == oldRelease.major
-                    && release.minor == oldRelease.minor
-                    && release.patch > oldRelease.patch
+    private func fetchNewToolchain(_ bounds: UpdateParameters) async throws -> ToolchainVersion? {
```
adam-fowler, line 152: Can we function header comments here and on other functions. Is this downloading the toolchain?

Sources/Swiftly/Update.swift
```diff
@@ -31,61 +38,154 @@ struct Update: SwiftlyCommand {
...
-            return nil
+        case let .snapshot(old):
+            return try await self.httpClient.getSnapshotToolchains(limit: 1) { snapshot in
+                snapshot.branch == old.branch && snapshot.date > old.date
+            }.first.map(ToolchainVersion.snapshot)
+        }
+    }
+}
+
+enum UpdateParameters {
```
adam-fowler, line 173: Again comments please

Sources/SwiftlyCore/HTTPClient.swift
adam-fowler: Can we add some comment above `ToolchainDownloader` detailing why it is needed

Sources/SwiftlyCore/HTTPClient.swift
```diff
@@ -117,43 +139,57 @@ public class HTTP {
...
-        }
+        if let downloader = self.downloader {
+            return try await downloader.downloadToolchain(
+                toolchain,
+                url: url,
+                to: destination,
+                reportProgress: reportProgress
+            )
+        } else {
+            let fileHandle = try FileHandle(forWritingTo: URL(fileURLWithPath: destination))
```
adam-fowler, line 156: Why is this not implemented as a ToolchainDownloader

adam-fowler (changes requested):

Sources/Swiftly/Update.swift
patrickfreed: Done

Sources/Swiftly/Use.swift
patrickfreed: The config variables being passed in were almost always named `config` themselves, so the call sites looked a little duplicative.

Sources/SwiftlyCore/HTTPClient.swift
patrickfreed: done

Sources/SwiftlyCore/HTTPClient.swift
patrickfreed: That's how it used to be, but it was changed to be this way so that tests could construct individual commands with their own mocked HTTP clients. The lifetime of a swiftly process is only a single command in actual usage, so having them per-command is essentially the same as a singleton in practice.

Sources/SwiftlyCore/HTTPClient.swift
patrickfreed: No good reason in particular, updated to be.

adam-fowler (approved):
I think we are good

PR #76, opened 2023-10-11
Add --use flag to install

### LINKED ISSUE #71: Provide option to use the newly installed toolchain
The `install` command should accept an argument (e.g. `--use`) which will automatically use the new toolchain if installation succeeds. This already happens by default if there are no toolchains installed yet, but this would be very useful in other cases too.

patrickfreed:
Closes #71

### DIFF
--- a/Sources/Swiftly/Update.swift
+++ b/Sources/Swiftly/Update.swift
@@ -100,11 +100,12 @@ struct Update: SwiftlyCommand {
             }
         }
 
-        try await Install.execute(version: newToolchain, &config, self.httpClient, useInstalledToolchain: false)
-
-        if config.inUse == parameters.oldToolchain {
-            try await Use.execute(newToolchain, &config)
-        }
+        try await Install.execute(
+            version: newToolchain,
+            &config,
+            self.httpClient,
+            useInstalledToolchain: config.inUse == parameters.oldToolchain
+        )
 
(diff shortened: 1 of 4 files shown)


Sources/Swiftly/Update.swift
```diff
@@ -56,7 +56,7 @@ struct Update: SwiftlyCommand {
         }
 
         print("updating \(oldToolchain) -> \(newToolchain)")
-        try await Install.execute(version: newToolchain)
+        try await Install.execute(version: newToolchain, use: false)
```
adam-fowler, line 59: I know update hasn't been merged yet but shouldn't `swiftly update` also default to using the newly updated toolchain.

adam-fowler (approved):
Looks good. One minor question.

Sources/Swiftly/Update.swift
patrickfreed: This is just in here to have the existing update code compile, but it'll be changed in #67. The planned logic for this is if the toolchain being updated is the one that is in use, then the newly installed one will be used, otherwise no.

patrickfreed:
Going to wait to merge this until #67 goes in so that I can reuse the download mocking for the test I added here. The RHEL 9 failure is due to a missing download on that platform, but we don't actually need to download anything for this test.

PR #81, opened 2023-11-15
Implement `self-update` command

patrickfreed:
Implements a `self-update` command, which can be used to update swiftly in-place.

This also expands the toolchain downloading mocking framework introduced in #67 to be able to be used for any arbitrary HTTP request. This allows us to use mocking to simulate there being a swiftly update to make, even though there are none currently.

### DIFF
--- a/Sources/SwiftlyCore/HTTPClient.swift
+++ b/Sources/SwiftlyCore/HTTPClient.swift
@@ -16,80 +16,64 @@ public protocol ToolchainDownloader {
     ) async throws
 }
 
-/// The default implementation of a toolchain downloader.
-/// Downloads toolchains from swift.org.
-private struct HTTPToolchainDownloader: ToolchainDownloader {
-    func downloadToolchain(
-        _: ToolchainVersion,
-        url: String,
-        to destination: String,
-        reportProgress: @escaping (SwiftlyHTTPClient.DownloadProgress) -> Void
-    ) async throws {
-        let fileHandle = try FileHandle(forWritingTo: URL(fileURLWithPath: destination))
-        defer {
-            try? fileHandle.close()
-        }
+public protocol HTTPRequestExecutor {
+    func execute(_ request: HTTPClientRequest, timeout: TimeAmount) async throws -> HTTPClientResponse
+}
 
-        let request = SwiftlyHTTPClient.client.makeRequest(url: url)
-        let response = try await SwiftlyHTTPClient.client.inner.execute(request, timeout: .seconds(30))
+private struct MockHTTPClient: HTTPRequestExecutor {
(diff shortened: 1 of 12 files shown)


Sources/SwiftlyCore/HTTPClient.swift
```diff
@@ -5,91 +5,49 @@ import NIO
...
 
-        try fileHandle.synchronize()
+    public func execute(_ request: HTTPClientRequest, timeout: TimeAmount) async throws -> HTTPClientResponse {
+        try await Self.client.inner.execute(request, timeout: timeout)
     }
 }
 
+private func makeRequest(url: String) -> HTTPClientRequest {
+    var request = HTTPClientRequest(url: url)
+    request.headers.add(name: "User-Agent", value: "swiftly")
```
adam-fowler, line 23: You should include the swiftly version here as well

Sources/SwiftlyCore/HTTPClient.swift
```diff
@@ -5,91 +5,49 @@ import NIO
...
 public struct SwiftlyHTTPClient {
-    fileprivate static let client = HTTPClientWrapper()
-
     private struct Response {
         let status: HTTPResponseStatus
         let buffer: ByteBuffer
     }
 
-    private let downloader: ToolchainDownloader
+    private let inner: HTTPRequestExecutor
```
adam-fowler, line 34: Rename to executor?

Sources/SwiftlyCore/HTTPClient.swift
```diff
@@ -179,30 +137,60 @@ public struct SwiftlyHTTPClient {
...
+
+        guard case response.status = HTTPResponseStatus.ok else {
+            throw Error(message: "Received \(response.status) when trying to download \(url)")
+        }
+
+        // Unknown download.swift.org paths redirect to a 404 page which then returns a 200 status.
+        // As a heuristic for if we've hit the 404 page, we check to see if the content is HTML.
+        if url.host == "download.swift.org" {
+            guard !response.headers["Content-Type"].contains(where: { $0.contains("text/html") }) else {
+                throw SwiftlyHTTPClient.DownloadNotFoundError(url: url.path)
```
adam-fowler, line 157: I was going to ask can't you check the response status, but apparently not. The 404 page returns a status of 302. You could check the location header is `https://swift.org/404.html` instead though

adam-fowler (approved):
In general looks good. Adding a few comments

Sources/SwiftlyCore/HTTPClient.swift
patrickfreed: So after following the redirect, swift.org/404.html does in fact return a status code of 404 nowadays, so I've just gone ahead and removed this special check altogether.

PR #86, opened 2023-12-10
Support updating exsting swiftly installation from swiftly-install.sh

### LINKED ISSUE #82: `swiftly-install.sh` doesn't clean up existing symlinks when performing a clean installation
When performing a clean installation via `swiftly-insall.sh`, any existing symlinks to swiftly-managed toolchains are not cleaned up, despite the pointed-to toolchain being deleted. This leaves a bunch of dangling symlinks around. We should ensure that these symlinks are cleaned up when a clean install is performed.

### LINKED ISSUE #83: Support upgrading the swiftly installation via `swiftly-install.sh`
`swiftly-install.sh` currently can only perform fresh installations. We should update it to support just updating the swiftly binary while leaving installed toolchains alone.

We may want to preserve the clean installation functionality via an argument.

patrickfreed:
Closes #83, also fixes #82

This PR updates the default behavior of swiftly-install.sh when an existing installation exists from prompting to perform a clean installation (i.e. delete all existing configuration and installed toolchains) to simply replacing the swiftly binary. Users can opt into the old behavior (without prompting) by using a newly introduced `--overwrite` option.

Going forward, most users will update swiftly via the `self-update` command introduced in #81, but since the current released swiftly version doesn't have that command, I think we need to provide a path through `swiftly-install.sh`. Furthermore, it would be too easy to accidentally delete your swiftly installation with the old default behavior, given that there isn't another supported means of upgrading.

Once this is done,  I think I'm good to go ahead with releasing 0.2.0 of swiftly (and 0.3.0 of swiftly-install.sh).

### DIFF
--- a/install/run-tests.sh
+++ b/install/run-tests.sh
@@ -17,19 +17,18 @@ fi
 
 tests_failed=0
 tests_passed=0
+failed_tests=()
 
 for t in tests/*.sh; do
+    test_name=$(basename "$t")
     line_print
-    echo "Running test $t"
+    echo "Running test $test_name"
     echo ""
     if bash "$t"; then
-        echo ""
-        echo "$t PASSED"
         ((tests_passed++))
     else
-        echo ""
-        echo "$t FAILED"
         ((tests_failed++))
+        failed_tests+=("$test_name")
     fi
 done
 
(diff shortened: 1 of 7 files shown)


patrickfreed:

patrickfreed:

patrickfreed:
The test failures look like they're unrelated--the ubuntu 18.04 machine used to run them might be out of disk space or something.

adam-fowler (approved):
Looks good

Issue #85, opened 2023-11-16
Prompt to remove existing Swift installation 


MahdiBM:
* Updated to 5.9.1 via Swiftly, `swift --version` would still show `5.8.1`.
* Tried `which swift` which said: `/usr/bin/swift`.
* I had noticed Swiftly says its installation is at `/root/.local/bin/swift` instead.
* Manually deleted `/usr/bin/swift`.
* Added `/root/.local/bin/` to PATH.
  * Not sure if this not-existing there wasn't my fault
  * For more context I overrided the older installation of Swiftly with a new one, not sure if it matters.
* `swift --version` now correctly shows `5.9.1`.

This process could be handled by Swiftly itself.
There could be a prompt to ask user to remove their existing Swift installation if it's in a non-Swiftly directory.

adam-fowler:
That sounds particularly dangerous asking Swiftly to remove a directory it doesn't manage.

patrickfreed:
I agree that removing the installation is probably not safe for us to do as part of swiftly's installation, but adding a warning that a different installation has precedence sounds like a decent idea to me.

Regarding adding it to your path, swiftly creates an `env.sh` in swiftly's home directory (which is by default `$HOME/.local/share/swiftly`) which updates your PATH correctly, and by default swiftly adds a line to source this file to your `.profile`. Next time you log in (or do `bash -login`) it should be picked up automatically, or you can just source it in your current shell.

swiftly-install.sh is also _supposed_ to print a message instructing you to source that file in your current shell if `$HOME/.local/bin` isn't already on your PATH, but it looks like if you have an existing installation of Swift elsewhere, this message isn't printed properly. We should definitely fix that.

carlynorama:
Had the same problem, but I just restarted the shell.  Offer up a temp fix of suggesting a manual path update in that same message that says to run env.sh

https://github.com/swift-server/swiftly/pull/93

patrickfreed:
Just pushed a commit (https://github.com/swift-server/swiftly/commit/76a23583dc177b369f9be54ab6f338bef37c77be) to print a warning when an existing installation has been detected:
```
swiftly has been succesfully installed!

Once you log in again, swiftly should be accessible from your PATH.
To begin using swiftly from your current shell, first run the following command:

    . $HOME/.local/share/swiftly/env.sh

Then to install the latest version of Swift, run 'swiftly install latest'

Warning: existing installation of Swift detected at /usr/bin/swift
[...]

PR #93, opened 2024-02-09
Update swiftly-install.sh

carlynorama:
Just installed swiftly on a machine with a version of Swift and noticed it didn't switch. I assumed it was a path problem, but was confused because I ran the script which I had assumed done that for me.  I just reloaded the shell and everything was fine.

I did notice issue #85 though, and thought I'd update the message to provide instructions on doing a manual update if that's still the preferred fix.  UPDATE: I think actually my problem was less I had a version of Swift installed as much as I forgot I had set the user's shell to zsh, but as others may do the same. For the record I did see that message.


patrickfreed:
I ended up addressing this when adding the message about existing Swift installations: https://github.com/swift-server/swiftly/commit/76a23583dc177b369f9be54ab6f338bef37c77be As part of that, I opted to print `hash -r` rather than `rehash`, since both zsh and bash support the former.

Going to close this out, but thank you for contributing nonetheless!

carlynorama:
Looks great!

PR #94, opened 2024-02-14
Verify signatures of toolchains before installing

### LINKED ISSUE #11: Validate toolchain signatures before installation
To ensure the toolchains installed by Swiftly are valid and authentic, we should validate their signatures. The processes for doing this on macOS and Linux are outlined here: https://www.swift.org/download/#using-downloads

patrickfreed:
Closes #11

### DIFF
--- a/Sources/LinuxPlatform/Linux.swift
+++ b/Sources/LinuxPlatform/Linux.swift
@@ -141,5 +141,58 @@ public struct Linux: Platform {
         FileManager.default.temporaryDirectory.appendingPathComponent("swiftly-\(UUID())")
     }
 
+    public func validateSignture(httpClient: SwiftlyHTTPClient, archiveDownloadURL: URL, archive: URL) async throws {
+        guard self.commandExists("gpg") else {
+            SwiftlyCore.print("gpg not installed, skipping signature validation.")
+            return
+        }
+
+        SwiftlyCore.print("Refreshing Swift PGP keys...")
+        do {
+            try self.runShellCommand("gpg",
+                                "--quiet",
+                                "--keyserver", "hkp://keyserver.ubuntu.com",
+                                "--refresh-keys", "Swift")
+        } catch {
+            throw Error(message: "Failed to refresh PGP keys: \(error)")
+        }
+
+        SwiftlyCore.print("Downloading toolchain signature...")
+        let sigFile = self.getTempFilePath()
+        FileManager.default.createFile(atPath: sigFile.path, contents: nil)
+        defer {
+            try? FileManager.default.removeItem(at: sigFile)
+        }
+
+        try await httpClient.downloadFile(
+            url: archiveDownloadURL.appendingPathExtension("sig"),
+            to: sigFile
+        )
+
+        SwiftlyCore.print("Validating toolchain signature...")
+        do {
+            try self.runShellCommand("gpg", "--verify", sigFile.path, archive.path)
+        } catch {
+            throw Error(message: "Toolchain signature validation failed: \(error)")
+        }
+    }
+
+    private func commandExists(_ program: String) -> Bool {
+        return (try? self.runShellCommand("command", "-v", program, "> /dev/null")) != nil
+    }
+
+    private func runShellCommand(_ command: String...) throws {
+        let process = Process()
+        process.executableURL = URL(fileURLWithPath: "/usr/bin/env")
+        process.arguments = ["sh", "-c"] + command
+
+        try process.run()
+        process.waitUntilExit()
+
+        guard process.terminationStatus == 0 else {
+            throw Error(message: "\(command.first!) exited with non-zero status: \(process.terminationStatus)")
+        }
+    }
+
     public static let currentPlatform: any Platform = Linux()
 }
--- a/Sources/LinuxPlatform/Linux.swift
+++ b/Sources/LinuxPlatform/Linux.swift
@@ -25,6 +25,49 @@ public struct Linux: Platform {
         true
     }
 
+    private static let skipVerificationMessage: String = "To skip signature verification, specify the --no-verify flag."
+
+    public func verifySystemPrerequisitesForInstall(requireSignatureValidation: Bool) throws {
+        // The only prerequisite at the moment is that gpg is installed and the Swift project's keys have been imported.
+        guard requireSignatureValidation else {
+            return
+        }
+
+        guard (try? self.runProgram("gpg", "--version", quiet: true)) != nil else {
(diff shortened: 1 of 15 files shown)


patrickfreed:

patrickfreed:

patrickfreed:
@swift-server-bot test please

patrickfreed:

patrickfreed:

Sources/LinuxPlatform/Linux.swift
```diff
@@ -141,5 +141,76 @@ public struct Linux: Platform {
...
+                "gpg",
+                "--quiet",
+                "--keyserver",
+                "hkp://keyserver.ubuntu.com",
+                "--refresh-keys",
+                "Swift"
+            )
+        } catch {
+            throw Error(message: "Failed to refresh PGP keys: \(error)")
+        }
```
patrickfreed, line 175: This takes about 4 seconds to perform, so we could consider only running it every X days or something like that to avoid having to do that on every install. On the other hand, in the event of a key revocation, we'd want to refresh them immediately to ensure any invalid keys aren't ever used. What do you think would be a good policy here?

Sources/LinuxPlatform/Linux.swift
```diff
@@ -141,5 +141,76 @@ public struct Linux: Platform {
...
     }
 
+    public func verifySignature(httpClient: SwiftlyHTTPClient, archiveDownloadURL: URL, archive: URL) async throws {
+        guard (try? self.runProgram("gpg", "--version", quiet: true)) != nil else {
+            SwiftlyCore.print("gpg not installed, skipping signature verification.")
+            return
+        }
+
+        let foundKeys = (try? self.runProgram(
+            "gpg",
```
patrickfreed, line 151: I opted to just call through to the `gpg` executable rather than use a library for simplicity, especially since users are going to need to have installed it anyways in order to import the keys.

Sources/Swiftly/Install.swift
```diff
@@ -56,25 +56,35 @@ struct Install: SwiftlyCommand {
     ))
     var token: String?
 
+    @Flag(help: "Skip PGP verification of the installed toolchain's signature.")
+    var noVerify = false
```
patrickfreed, line 60: While this will add a new step to installation by default, I figured it would be best to default to best practices/security. As a compromise, if `gpg` isn't installed or the keys haven't been imported, verification will just be skipped to avoid breaking any existing users. I do wonder if we should also have a `--verify` argument that will proceed with installation if and only if verification passes. WDYT?

adam-fowler: I think the default should be to fail if `gpg` isn't installed, or the verification fails. At both points we should ask the user if they want to continue with the default being "no". There is already a `--no-verify` argument to avoid this. We could also assume `--y` is the equivalent to `--no-verify`

I see swiftly does not import the keys for the user. Is this something it could do?

adam-fowler, line 60: Having an argument called `noVerify` means you have double negatives all over the place. Replace it with this and you get a `--no_verify` argument, as well as the `--verify`
```swift
    @Flag(inversion: .prefixedNo, help: "Perform PGP verification of the installed toolchain's signature.")
    var noVerify = false
```

Sources/LinuxPlatform/Linux.swift
```diff
@@ -141,5 +141,76 @@ public struct Linux: Platform {
...
+            "--list-keys",
+            "swift-infrastructure@forums.swift.org",
+            "swift-infrastructure@swift.org",
+            quiet: true
+        )) != nil
+        guard foundKeys else {
+            SwiftlyCore.print("Swift PGP keys not imported, skipping signature verification.")
```
adam-fowler, line 158: Ask do you want to skip? Defaulting to "no"

Sources/LinuxPlatform/Linux.swift
```diff
@@ -141,5 +141,76 @@ public struct Linux: Platform {
...
+            SwiftlyCore.print("To enable verification, import the keys from https://swift.org/keys/all-keys.asc")
```
adam-fowler, line 159: Can we provide help on how to do that. eg Provide `wget -q -O - https://swift.org/keys/all-keys.asc |
  gpg --import -`

Sources/Swiftly/Install.swift
adam-fowler: I'd rather get this right than worry too much about breaking existing users. This is v0.x still

Sources/LinuxPlatform/Linux.swift
patrickfreed: Since we have the `--no-verify` option, I'd rather just use that for any optionality related to verification, if that's okay. This let's us avoid additional interactivity in the install path, which is nice for using swiftly in CI environments without an active terminal user.

Sources/LinuxPlatform/Linux.swift
patrickfreed: Good idea, done.

Sources/Swiftly/Install.swift
patrickfreed: >I think the default should be to fail if gpg isn't installed, or the verification fails. At both points we should ask the user if they want to continue with the default being "no". There is already a --no-verify argument to avoid this. We could also assume --y is the equivalent to --no-verify

I agree that failing in these cases is probably better, updated to do so. That said, I'd like to avoid the interactive step if possible (see my comment above). Also, `install` doesn't currently have a `-y` flag as the command doesn't have any interactive prompts.

>I see swiftly does not import the keys for the user. Is this something it could do?

`swiftly-install.sh` will do that for users, but swiftly doesn't currently. If possible, I'd prefer to keep all non-toolchain installation tasks limited to the install script to avoid having to duplicate the logic for it / separate the concerns better. If this ends up being a pain point for users I'm definitely happy to revisit the idea, though I think it should be fine since most users will get the keys during install.

patrickfreed: good idea, updated to this.

adam-fowler:
If you don't have any of the requirements to check the toolchain signature, you get an error message after downloading the tar ball. You do as suggested and then have to download the tar ball again. Perhaps we should verify for the existence of `gpg` and that the swift PGP keys have been imported before we do the download.

adam-fowler:
Also RHEL and AL2 are failing

patrickfreed:

adam-fowler (approved):

adam-fowler:
Yeah Swift have stopped supporting 18.04
https://forums.swift.org/t/dropping-support-for-ubuntu-18-04/69561

PR #115, opened 2024-06-12
Ubuntu 24.04, 23.10 Fedora 39 and Debian 12

adam-fowler:
Add support in install script for Ubuntu 24.04, 23.10 Fedora 39 and Debian 12
Also use Dockerfiles from swift-docker/swift-ci/master if Dockerfile does not exist in nightly-main
Replace calls to printf with echo as on Fedora dockerfile has `%{` which is treated as a formatted parameter

### DIFF
--- a/install/swiftly-install.sh
+++ b/install/swiftly-install.sh
@@ -232,6 +232,14 @@ detect_platform () {
 
         *"ubuntu"*)
             case "$UBUNTU_CODENAME" in
+                "noble")
+                    set_platform_ubuntu "24" "04"
+                    ;;
+
+                "mantic")
+                    set_platform_ubuntu "23" "10"
+                    ;;
+
                 "jammy")
                     set_platform_ubuntu "22" "04"
                     ;;
@@ -274,31 +282,41 @@ manually_select_platform () {
     echo "Please select the platform to use for toolchain downloads:"
 
     echo "0) Cancel"
-    echo "1) Ubuntu 22.04"
-    echo "2) Ubuntu 20.04"
-    echo "3) Ubuntu 18.04"
-    echo "4) RHEL 9"
-    echo "5) Amazon Linux 2"
+    echo "1) Ubuntu 24.04"
+    echo "2) Ubuntu 23.10"
+    echo "3) Ubuntu 22.04"
+    echo "4) Ubuntu 20.04"
+    echo "5) Ubuntu 18.04"
+    echo "6) RHEL 9"
+    echo "7) Amazon Linux 2"
 
     read_input_with_default "0"
     case "$READ_INPUT_RETURN" in
         "1" | "1)")
-            set_platform_ubuntu "22" "04"
+            set_platform_ubuntu "24" "04"
             ;;
 
         "2" | "2)")
-            set_platform_ubuntu "20" "04"
+            set_platform_ubuntu "23" "10"
             ;;
 
         "3" | "3)")
-            set_platform_ubuntu "18" "04"
+            set_platform_ubuntu "22" "04"
             ;;
 
         "4" | "4)")
-            set_platform_rhel "9"
+            set_platform_ubuntu "20" "04"
             ;;
 
         "5" | "5)")
+            set_platform_ubuntu "18" "04"
+            ;;
+
+        "6" | "6)")
+            set_platform_rhel "9"
+            ;;
+
+        "7" | "7)")
             set_platform_amazonlinux "2"
             ;;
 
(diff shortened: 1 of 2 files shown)


adam-fowler:
@shahmishal what is needed done here to CI working for swiftly

futurejones:
> Also use Dockerfiles from swift-docker/swift-ci/master as these are more up to date

These Dockerfiles are for building the swift toolchain from source and include many extra dependencies not required for a normal swift installation.

adam-fowler:
> > Also use Dockerfiles from swift-docker/swift-ci/master as these are more up to date
> 
> These Dockerfiles are for building the swift toolchain from source and include many extra dependencies not required for a normal swift installation.

Currently swiftly-install.sh parses the Dockerfiles to extract the dependencies required to run swift, not an ideal solution. I'm hoping we can get some files served from `swift.org` with these details at some point. I guess if the swift-ci files are for building swift this isn't correct. I chose these because docker files for the new OSs did not exist at the top level. I change this to load from the original nightly-main and use swift-ci as a backup if files don't exist in nightly-main

adam-fowler:
@patrickfreed

patrickfreed:
Thanks for fixing this up--will take a look tonight.

0xTim:
See https://github.com/apple/swift-org-website/pull/720 for reference

install/swiftly-install.sh
```diff
@@ -274,34 +332,54 @@ manually_select_platform () {
     echo "Please select the platform to use for toolchain downloads:"
 
     echo "0) Cancel"
-    echo "1) Ubuntu 22.04"
```
patrickfreed, line 277: I think we should update this selection to also allow the user to specify the actual platform string (e.g. "Ubuntu 20.04"). That way reordering the selections like this won't break any user scripts that might rely on a number being associated with a particular platform.

install/swiftly-install.sh
```diff
@@ -127,29 +127,39 @@ install_system_deps () {
...
 
-    dockerfile_url="https://raw.githubusercontent.com/apple/swift-docker/main/nightly-main/$docker_platform_name/$docker_platform_version/Dockerfile"
+    dockerfile_url="https://raw.githubusercontent.com/swiftlang/swift-docker/main/nightly-main/$docker_platform_name/$docker_platform_version/Dockerfile"
+    set +e
     dockerfile="$(curl --silent --retry 3 --location --fail $dockerfile_url)"
-    if [[ "$?" -ne 0 ]]; then
+    # if we couldn't find Dockerfile in nightly-main use swift-ci/master version
+    if [[ -z "$dockerfile" ]]; then
+        dockerfile_url="https://raw.githubusercontent.com/swiftlang/swift-docker/main/swift-ci/master/$docker_platform_name/$docker_platform_version/Dockerfile"
+        dockerfile="$(curl --silent --retry 3 --location --fail $dockerfile_url)"
```
patrickfreed, line 136: to avoid duplicating this incantation, can we factor it out to a function we can call here and above? we can also move the `set +e` and `set -e` in there too.

patrickfreed (commented):
looks great, just have two tiny suggestions.

Looks like we're going to need to update the testing matrix to include these new platforms. @yim-lee is that something you can help with (or @shahmishal)?

yim-lee:
> @yim-lee is that something you can help with

@patrickfreed Sorry nope, I don't have the power here.

install/swiftly-install.sh
adam-fowler: Do you want the curl call and `set +e -e` in a separate function or the full test one location if it doesn't exist test the other location

install/swiftly-install.sh
adam-fowler: Can't people already use `-p` to be sure of what platform to get

patrickfreed: Ah good point, yeah this is fine then.

install/swiftly-install.sh
patrickfreed: I was just thinking the curl and `set +e -e` so we could avoid duplicating the `curl` invocation. That way if we ever need to change the arguments or whatever we can do it in once place.

shahmishal:

shahmishal:

adam-fowler:

patrickfreed:
@shahmishal any update on getting these new platforms running in CI?

MahdiBM:
Worth noting, now that [Apple is opening up to GitHub Actions](https://github.com/apple/swift-nio/pull/2760), we could consider moving to GitHub Actions as well, sometime in the future.

patrickfreed:
Oh moving to GitHub actions would be wonderful. Would that be an option for swiftly @shahmishal?

vsarunas:
Would be great to merge this PR.

adam-fowler:
This hasn't been merged because the init phase of swiftly is being rewritten to be implemented inside swiftly itself. We'll make sure to include the new OS's then.

vsarunas:
Thanks,  I assume in https://github.com/swiftlang/swiftly/pull/127 ?

adam-fowler:
> Thanks, I assume in #127 ?

yes

adam-fowler:
Closing in favour of #152

PR #116, opened 2024-06-12
Swiftly Enhancements - macOS Support, Command Proxies, and Shell Script Migration

cmcgee1024:
This pull request rolls up enhancements in several commits (more details are available in the commit messages) in the following areas:

## macOS Support
Support installing toolchains on macOS from the .pkg files available on swift.org. While the toolchains install in a different directory, the swiftly interactions remain much the same.

## Command Proxies
Instead of global symbolic links, Swiftly installs a set of proxies that can be configured to route to a particular toolchain using a number of criteria in order of precedence. Swiftly knows which tool is being proxies based on the `argv[0]` of the invocation. If the user provides a selector using a special '+' prefix to the argument then Swiftly will use a matching toolchain with the highest precedence. Secondly, the next highest precedence is the presence of a `.swift-version` file that provides a toolchain version to use. Finally, there can be a global default toolchain, which is often the very first one installed using Swiftly, or using the `swiftly use` command.

The swiftly use subcommand now updates or installs a `.swift-version` to the version of the toolchain that is used. There is a global option that overrides this behaviour and updates the global default in the configuration.

## Shell Script Migration to Swiftly
The shell script is not longer required to install Swiftly. Instead, Swiftly itself, when invoked with `argv[0]` as `swiftly-init` or swiftly run with the init sub-command, will begin the installation process. This process was migrated from the shell script with one user visible change to the procedure, which is that system packages are not automatically installed. Instead, the presence of the system packages is detected, one-by-one, and the user is given a command that they should run as root via sudo or some means so that they can successfully run the tools in the toolchain.

There are some other items covered by this PR:
* Updates to the README and RELEASING files with details about how to install Swiftly for end users, uninstall older versions of Swiftly to prepare for the new release, and invoke the different test suites
* Remove unsafe byte buffer management code in SwiftlyHTTPClient for better security
* Upgrade libarchive to take advantage of improvements, including some security fixes
* Add a build script for macOS in preparation for providing macOS packages on swift.org
* Change the build script for Linux to generate GPG signatures and match the usual tar.gz assets on swift.org
* Small bug fixes to the docker compose files that doubled the double quotes for pretty names during testing


patrickfreed:
Super excited to see all these changes! I personally haven't had a chance to look through them all yet, but I had a few preliminary questions/notes:
- it might be better to split this PR up into a few distinct ones, since each of the individual changes are quite significant on their own and would benefit from individual review/testing.
- One of the motivations behind having a separate bash script for installing swiftly itself was to enable swiftly to be installed and completed set up with a single bash one-liner. Now that swiftly-install.sh has been reimplemented in Swift, do we plan to maintain that (I personally would advocate for doing so at least :slightly_smiling_face:)? I think we'd be able to achieve that with a far simpler `swiftly-install.sh` than we had before these changes, which basically would just download the correct swiftly binary (i.e. the ARM or x86_64 one) and then run `swiftly run init` on it.

adam-fowler:
Wow, that's a lot of work. But we can't merge this in its current form. The individual components need to be discussed separately. I would suggest we attempt to get the macOS support in first.

FranzBusch:
I agree with @adam-fowler here. Can we get this split into individual PRs that focus on one thing at a time?

cmcgee1024:
Thanks everyone for having a look at this. I'm going to close it and start opening smaller and more focused PR's.

cmcgee1024:
> * One of the motivations behind having a separate bash script for installing swiftly itself was to enable swiftly to be installed and completed set up with a single bash one-liner. Now that swiftly-install.sh has been reimplemented in Swift, do we plan to maintain that (I personally would advocate for doing so at least 🙂)? I think we'd be able to achieve that with a far simpler `swiftly-install.sh` than we had before these changes, which basically would just download the correct swiftly binary (i.e. the ARM or x86_64 one) and then run `swiftly run init` on it.

Thanks for the background on the intent of the original design and the suggestions.

There were some reasons that I wanted to migrate the installation logic in swiftly itself instead of the shell script. The ergonomics of maintaining a shell script as a separate versionable piece that works across platforms is awfully painful. You'll see in the macOS PR I just raised some of the hoops and ugly shell conditional logic to work around getopt being more limited there than with the latest GNU getopt. :(

With Swift, we can get better platform abstractions, bring along rich argument parsing, and tap into the rich functionality available with better safety mechanisms. Also, with a single static binary, there are more packaging and delivery options, such as homebrew, or direct download from swift.org. I think that if the original goal was a shell one-liner we can certainly continue to do that. Maybe we can do something a little more transparent? I was thinking something like this that can be easily copy-pasted from swift.org with a little user agent platform detection.

```
curl https://swift.org/downloads/swiftly-0.4.0-RELEASE-linux-aarch64.tar.gz | (cd $(mktemp -d); tar -zxf -; ./swiftly-0.4.0-RELEASE-linux-aarch64/usr/bin/swiftly-init)
```

If the ergonomics don't feel right then perhaps we can offer a swiftly-install.sh script on swift.org that does effectively the same kind of thing, changing very rarely, extremely simple.

PR #118, opened 2024-06-13
Copy proper install command on copy button click

Frizlab:
The old URL was used in the install command instead of the new one.


adam-fowler (approved):

adam-fowler:
thank you

PR #120, opened 2024-06-14
Update usage of unsafe API, upgrade libarchive to 3.7.4, upgrade async http client

cmcgee1024:
Reduce the use of unsafe swift constructs in favour of ones that are memory safe. In the HTTP client remove the use of the unsafe mutable pointer in favour of direct Data conversion before writing the downloaded file to disk.

Upgrade the version of libarchive to 3.7.4 with various fixes.

Upgrade the async http client to fix an issue with decompression limits encountered with the GitHub REST API's.

### DIFF
--- a/Sources/SwiftlyCore/HTTPClient.swift
+++ b/Sources/SwiftlyCore/HTTPClient.swift
@@ -11,10 +11,8 @@ public protocol HTTPRequestExecutor {
 
 /// An `HTTPRequestExecutor` backed by an `HTTPClient`.
 internal struct HTTPRequestExecutorImpl: HTTPRequestExecutor {
-    fileprivate static let client = HTTPClientWrapper()
-
     public func execute(_ request: HTTPClientRequest, timeout: TimeAmount) async throws -> HTTPClientResponse {
-        try await Self.client.inner.execute(request, timeout: timeout)
+        try await HTTPClient.shared.execute(request, timeout: timeout)
     }
 }
 
@@ -166,9 +164,7 @@ public struct SwiftlyHTTPClient {
         for try await buffer in response.body {
             receivedBytes += buffer.readableBytes
 
-            try buffer.withUnsafeReadableBytes { bufferPtr in
-                try fileHandle.write(contentsOf: bufferPtr)
-            }
+            try fileHandle.write(contentsOf: buffer.readableBytesView)
 
             let now = Date()
             if let reportProgress, lastUpdate.distance(to: now) > 0.25 || receivedBytes == expectedBytes {
@@ -182,8 +178,4 @@ public struct SwiftlyHTTPClient {
 
         try fileHandle.synchronize()
     }
-}
-
-private class HTTPClientWrapper {
-    fileprivate let inner = HTTPClient.shared
 }
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,12 +1,13 @@
 {
+  "originHash" : "4e45e73ff02865a9bd0426ba4c6cd8dfcb967e62e7663128dd262908109e1487",
   "pins" : [
     {
       "identity" : "async-http-client",
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/swift-server/async-http-client",
       "state" : {
-        "revision" : "fb308ee72f3d4c082a507033f94afa7395963ef3",
-        "version" : "1.21.0"
+        "revision" : "0ae99db85b2b9d1e79b362bd31fd1ffe492f7c47",
+        "version" : "1.21.2"
       }
     },
     {
@@ -136,5 +137,5 @@
       }
     }
   ],
-  "version" : 2
+  "version" : 3
 }
--- a/Package.swift
+++ b/Package.swift
@@ -13,7 +13,7 @@ let package = Package(
     ],
     dependencies: [
         .package(url: "https://github.com/apple/swift-argument-parser", from: "1.3.0"),
-        .package(url: "https://github.com/swift-server/async-http-client", from: "1.21.0"),
+        .package(url: "https://github.com/swift-server/async-http-client", from: "1.21.2"),
         .package(url: "https://github.com/apple/swift-nio.git", from: "2.64.0"),
         .package(url: "https://github.com/apple/swift-tools-support-core.git", from: "0.6.1"),
     ],
--- a/scripts/install-libarchive.sh
+++ b/scripts/install-libarchive.sh
@@ -3,7 +3,7 @@
 set -o errexit
 
 # TODO detect platform
-LIBARCHIVE_VERSION=3.6.1
+LIBARCHIVE_VERSION=3.7.4
 
 mkdir /tmp/archive-build
 pushd /tmp/archive-build


Sources/SwiftlyCore/HTTPClient.swift
```diff
@@ -166,8 +166,9 @@ public struct SwiftlyHTTPClient {
         for try await buffer in response.body {
             receivedBytes += buffer.readableBytes
 
-            try buffer.withUnsafeReadableBytes { bufferPtr in
-                try fileHandle.write(contentsOf: bufferPtr)
+            let byteData = buffer.getData(at: buffer.readerIndex, length: buffer.readableBytes)
+            if let data = byteData {
+                try fileHandle.write(contentsOf: data)
```
adam-fowler, line 171: Can probably replace all of this with 
```swift
try fileHandle.write(contentsOf: buffer.readableBytesView)
```

adam-fowler (approved):
Hi Chris, welcome onboard
This looks good, but I've a suggested an alternative.

Sources/SwiftlyCore/HTTPClient.swift
cmcgee1024: Thank you for the suggestion, this is much cleaner. I've pushed an update that uses it.

Sources/SwiftlyCore/HTTPClient.swift
```diff
@@ -185,5 +183,5 @@ public struct SwiftlyHTTPClient {
 }
 
 private class HTTPClientWrapper {
-    fileprivate let inner = HTTPClient.shared
+    fileprivate let inner = HTTPClient(eventLoopGroupProvider: .singleton)
```
cmcgee1024, line 186: Note that this was added because in ad hoc testing I discovered an emergent http decompression limit error. It appears that something has changed to trigger it in the GitHub API's.

adam-fowler: Does `.shared` have different limits to the HTTPClient with a default configuration?

cmcgee1024: It appears so. I dug a little deeper because I was concerned that this could be affecting everyone using swiftly. This is the commit that regressed and the testing didn't pick it up at the time: 2d054143ea29e2e5ee0085cbb0ca2ed8086d9166

I'm fully reverting that change so that this is not broken, but it does leave some open questions about our commit verification process. Ideally, some form integration testing might have picked this up at the time. Also, it does seem suspect that the shared HTTP Client has a broken decompression limit, whereas this initializer that just reuses the singleton event loop group provider has an intact decompression.

adam-fowler: Have you reported this in the async-http-client repository?

adam-fowler: Including @weissi as I believe this was from his PR to use shared

cmcgee1024: I can try and create a small reproducible test case and see if the latest fixes this. I hope to do that once I've restored this to a working state here with the original working code.

cmcgee1024: @adam-fowler the latest commit fully reverts the HTTPClient.swift portion of 2d0514 and smoke testing seems to be working again. Any chance that we can get these fixes in to get us back into a working state?

gwynne: We've run into this problem over in Penny (Vapor's Discord bot) - this was the fix Mahdi implemented: https://github.com/vapor/penny-bot/blob/main/Sources/Shared/%2BHTTPClient.swift#L12-L14

MahdiBM: Yes, and there is already an issue for it (which I should have marked as completed, but forgot):
https://github.com/swift-server/async-http-client/issues/739

@cmcgee1024 feel free to comment there if you think the limit is still too low for Swiftly, although I'm not sure how much success you'll have with getting the team to agree on increasing it, considering increasing the limit has some implications.

cmcgee1024: Thanks @gwynne 

I think that unless there's a reason why we can't revert the HTTPClient change here, we should probably go with that for now until this issue is fixed in async-http-client: https://github.com/swift-server/async-http-client/issues/739

adam-fowler: Having investigated, the reason `HTTPClient(eventLoopGroupProvider: .singleton)` works and `HTTPClient.shared` doesn't is the first disables decompression. I would rather we still ran with decompression enabled. If the current rate is too low then let's do what @MahdiBM suggests. I would also like this implemented as a singleton so we don't have to worry about shutdown.

Joannis: @adam-fowler heads up: you've approved this PR, but left a change request in this comment afterwards.

patrickfreed: >Ideally, some form integration testing might have picked this up at the time.

It appears maybe this was some change on the GitHub side that introduced this issue. I've just tried to run a basic install locally and hit the error, so I'd imagine if we had CI set up right now we'd see it. But yeah, it does indicate that we should be running tests on some regular interval (daily?) to ensure things like this can be caught as they happen.

patrickfreed:
I just looked into the background for the libarchive security issue, and in some ways it's pretty serious as it's related to the known bad actor that compromised the xz library. See https://github.com/libarchive/libarchive/pull/2101 and https://github.com/libarchive/libarchive/pull/1609 for more context.

That being said, it looks like no CVE was issued for it, and the suspicious code changes only affected the bsdtar utility, rather than the library functions as used by swiftly. So in light of that, I don't believe there are security concerns here for us. We should still proceed with the upgrade, but I don't think we should indicate that doing so is a security fix in the commit message or release notes, as that has certain implications for downstream users.

Likewise, I agree that moving away from using unsafe APIs in swiftly is an improvement that we should merge, but the case identified here was a safe usage of it, so there was no memory safety violation that we're now fixing. In the commit message / release notes, we should note this accordingly (e.g. "stop using unsafe API" rather than "memory safety fixes"). Again, the only reason I'm harping on this is security issues have to be handled with special care, and so if we did have an identified security problem, we would want to cut a release as soon as possible and notify users in the changelog what the case we're fixing is. Thankfully, that doesn't appear to be the case here though.

All that being said, changes look good to me mod @adam-fowler's comment about using a singleton.

patrickfreed:
I took the liberty of updating the PR title, let me know if it seems alright with you.

Sources/SwiftlyCore/HTTPClient.swift
FranzBusch: I agree with @adam-fowler here let's define a static one so we don't need the wrapper with the `deinit`

cmcgee1024: > It appears maybe this was some change on the GitHub side that introduced this issue. I've just tried to run a basic install locally and hit the error, so I'd imagine if we had CI set up right now we'd see it. But yeah, it does indicate that we should be running tests on some regular interval (daily?) to ensure things like this can be caught as they happen.

After thinking on this a bit, I don't think that GitHub is the variable as I had initially thought. Everything was working for me until I rebased this change on top of the above commit that changed the way that HTTPClient was initialized. A CI with at least a smoke test that did a `swiftly install latest` would have picked up the problem.

cmcgee1024: >  I would rather we still ran with decompression enabled. If the current rate is too low then let's create a client with a higher rate. I would also like this implemented as a singleton so we don't have to worry about shutdown. eg

Is this blocking the approval and merge (I don't merge permissions here) of this PR? Can we defer this to a new issue/PR? It appears that the existing approach before the regression was working for quite some time and is currently working for everyone on v0.3.0. I'm just restoring that so that we're all unblocked.

adam-fowler: > Is this blocking the approval and merge (I don't merge permissions here) of this PR? Can we defer this to a new issue/PR?

I would rather we stick to a singleton. The change I suggested works, as I used it when testing the macOS support PR

weissi: When you say 'a singleton', you mean `HTTPClient.shared` or your own singleton?

I very much think it should be `HTTPClient.shared`, there's no reason to create singletons all over the place, just wastes resources. The design goal is that `HTTPClient.shared` behaves like the platform browser: If Safari/Chrome/... can open it in defaults settings then `HTTPClient.shared` should do.

adam-fowler: @weissi I understand that but we cannot use `HTTPClient.shared` at the moment. It just keeps throwing errors about compression limits.

cmcgee1024: @adam-fowler I've updated this PR with your suggested change to the construction of the http client.

weissi: @adam-fowler Correct right this instant but the fix for this is merged, I'm working with the maintainers on cutting a release

weissi: Thanks to @lukasa, we've got https://github.com/swift-server/async-http-client/releases/tag/1.21.2 now. So `from: "1.21.2"` for AHC should fix this.

patrickfreed: @weissi / @Lukasa thanks for pushing that through!

>After thinking on this a bit, I don't think that GitHub is the variable as I had initially thought. Everything was working for me until I rebased this change on top of the above commit that changed the way that HTTPClient was initialized. A CI with at least a smoke test that did a swiftly install latest would have picked up the problem.

Prior to the move to swiftlang, we did run tests in CI, a few of which covered that case (`testInstallLatest` and `default-install.sh` in particular). That's what leads me to believe maybe something changed on the GitHub side. @shahmishal is working on getting CI set up under the new organization, so we'll have automated testing back shortly.

patrickfreed (approved):

shahmishal:

shahmishal:

shahmishal:

shahmishal:

shahmishal:

cmcgee1024:
Thanks everyone for your help with this.

PR #121, opened 2024-06-14
macOS support for swiftly

cmcgee1024:
It works much like it does already for Linux with some notable differences:

  * The toolchains are installed using the pkg files and macOS installer
  * The toolchain directory is ~/Library/Developer/Toolchains instead of ~/.local/share/swiftly/toolchains
  * The swiftly shared directory is ~/Library/Application Support/swiftly as it this is a more typical place for macOS applications to store their supporting files

Create a MacOS struct that implements the existing Platform protocol. Make a platform-specific target for this module. Bump the required swift toolchain version to resolve compiler errors and set the minimum macOS version to 13.

Update the README.md with some macOS details and fix some of the details that were outdated for Linux.

Add some helpful notes regarding the need to rehash the zsh on macOS since even when the swiftly bin directory has higher precedence in the PATH it sometimes gets snagged on the /usr/bin/swift, which doesn't detect the user installed toolchains and sometimes tries to get the user to install Xcode.

Make the shell script swiftly installer capable of operating in a standard macOS environment. First, detect that the environment is macOS, and then adjust the getopts for macOS's more limited implementation with the short opts. Also, remove any of the Linux specific steps to detect the distribution, check for gpg, and attempt to install Linux system packages.


Sources/MacOSPlatform/MacOS.swift
```diff
@@ -0,0 +1,184 @@
...
+        // All system prerequisites should be there for macOS
+    }
+
+    public func install(from tmpFile: URL, version: ToolchainVersion) throws {
+        guard tmpFile.fileExists() else {
+            throw Error(message: "\(tmpFile) doesn't exist")
+        }
+
+        if !self.swiftlyToolchainsDir.fileExists() {
+            try FileManager.default.createDirectory(at: self.swiftlyToolchainsDir, withIntermediateDirectories: false)
```
adam-fowler, line 36: Is this necessary given the installer will be creating this folder anyway

install/swiftly-install.sh
```diff
@@ -330,18 +330,27 @@ if ! has_command "curl" ; then
     exit 1
 fi
 
-if ! verify_getopt_install ; then
-    echo "Error: getopt must be installed from the util-linux package to run swiftly-install"
-    exit 1
+IS_MACOS="false"
+if has_command "sw_vers"; then
```
adam-fowler, line 334: Interesting way to recognise a Mac, is this standard?

install/swiftly-install.sh
```diff
@@ -356,20 +365,24 @@ USAGE:
...
+    -o, --overwrite             Overwrite the existing swiftly installation found at the configured
                                 SWIFTLY_HOME, if any. If this option is unspecified and an existing
                                 installation is found, the swiftly executable will be updated, but
                                 the rest of the installation will not be modified.
     -h, --help                  Prints help information.
-    --version                   Prints version information.
+    -v, --version               Prints version information.
+
+NOTES:
+    macOS only works with the short options (e.g. -v and not --version).
```
adam-fowler, line 384: Is this a limitation in macOS bash?

install/swiftly-install.sh
```diff
@@ -566,19 +585,27 @@ if [[ -f "$HOME_DIR/config.json" ]]; then
     detected_existing_installation="true"
     if [[ "$overwrite_existing_intallation" == "true" ]]; then
         echo "Overwriting existing swiftly installation at $(replace_home_path $HOME_DIR)"
```
adam-fowler, line 587: $HOME_DIR still seems to be pointing to $HOME/.local/share/swiftly and not $HOME/Library/Application Support/swiftly. See line 533

DESIGN.md
```diff
@@ -54,28 +54,23 @@ This is all very similar to how rustup does things, but I figure there's no need
 ## macOS
 ### Installation of swiftly
 
-Similar to Linux, the bootstrapping script for macOS will create a `~/.swiftly/bin` directory and drop the swiftly executable in it. A similar `~/.swiftly/env` file will be created and a message will be printed suggesting users add source `~/.swiftly/env` to their `.bash_profile` or `.zshrc`.
+Similar to Linux, the bootstrapping script for macOS will create a `~/.local/bin` directory and drop the swiftly executable in it. A `~/.local/share/swiftly/env` file will be created and a message will be printed suggesting users add source `~/.local/share/swiftly/env` to their `.bash_profile` or `.zshrc`.
```
adam-fowler, line 57: ~/Library/Application Support/swift?

adam-fowler (commented):
In general looks good.
We need to get the tests working though.

adam-fowler:
I don't want to push people into a position where they need Xcode installed to run with Swift (in the future I do hope we can install the swift toolchain without requiring Xcode). But I'm wondering, if Xcode is installed or at least `xcode-select` is available we use that for selecting the toolchain instead of pointing PATH to a folder of symbolic links. This would avoid the issue of clashing with /usr/bin/swift and work more in-sync with how Xcode works. Although this might mean we can't implement local proxies so I guess could be a deadend.

README.md
```diff
@@ -41,7 +41,7 @@ Target: x86_64-unknown-linux-gnu
 - Linux-based platforms listed on https://swift.org/download
   - CentOS 7 will not be supported due to some dependencies of swiftly not supporting it, however.
 
-Right now, swiftly is in the very early stages of development and is only supported on Linux, but the long term plan is to also support macOS. For more detailed information about swiftly's intended features and implementation, check out the [design document](DESIGN.md).
+Right now, swiftly is in early stages of development and is only supported on Linux and macOS. For more detailed information about swiftly's intended features and implementation, check out the [design document](DESIGN.md).
```
Joannis, line 44:
```suggestion
Right now, swiftly is in early stages of development and is supported on Linux and macOS. For more detailed information about swiftly's intended features and implementation, check out the [design document](DESIGN.md).
```

I think we can drop the "only" now. We're missing Windows, but certainly cover a lot more of the userbase now.

Sources/SwiftlyCore/Platform.swift
```diff
@@ -88,7 +88,12 @@ extension Platform {
 
     /// The "toolchains" subdirectory of swiftly's home directory. Contains the Swift toolchains managed by swiftly.
     public var swiftlyToolchainsDir: URL {
+        #if !os(macOS)
```
Joannis, line 91: Can we invert this?

```swift
#if os(macOS)

#elseif os(Linux)

#else
  #error("Unsupported platform")
#endif
```

I find that generally cognitively easier to parse, but also assume we'll be adding Windows at some point.

install/swiftly-install.sh
Joannis: Also curious about this

patrickfreed:
>In general looks good. I assume adding $HOME/.local/bin to the PATH is done in the install script. I can't remember offhand.

Yeah, by default the install script will add a line to source `$SWIFTLY_HOME_DIR/env.sh` to the system's preferred login script (typically `.profile`). `env.sh` is what actually adds `$SWIFTLY_BIN_DIR` to the PATH, which is by default `$HOME/.local/bin`.

This is the contents of my `env.sh` for example:
```bash
export SWIFTLY_HOME_DIR="$HOME/.local/share/swiftly"
export SWIFTLY_BIN_DIR="$HOME/.local/bin"
if [[ ":$PATH:" != *":$SWIFTLY_BIN_DIR:"* ]]; then
   export PATH="$SWIFTLY_BIN_DIR:$PATH"
fi
```

DESIGN.md
```diff
@@ -54,28 +54,23 @@ This is all very similar to how rustup does things, but I figure there's no need
...
+Instead of downloading tarballs containing the toolchains and storing them directly in `~/.local/share/swiftly/toolchains`, we instead install Swift toolchains to `~/Library/Developer/Toolchains` via the `.pkg` files provided for download at swift.org. (Side note: we’ll need to request that other versions than the latest be made available). To select a toolchain for use, we update the symlinks at `~/.local/bin` to point to the desired toolchain in `~/Library/Developer/Toolchains`. In the env file, we’ll contain a line that looks like `export PATH=”$HOME/.local/bin:$PATH`, so the version of swift being used will automatically always be from the active toolchain. `config.json` will contain version information about the selected toolchain as well as its actual location on disk.
```
FranzBusch, line 73: This side note is interesting. Are we working on making other toolchains available already? I would opt for removing this side note because if we make them available this side note is stale

patrickfreed: I think this side is stale. I just tried a few older releases and the downloads are available, e.g.
```
https://download.swift.org/swift-5.9.2-release/xcode/swift-5.9.2-RELEASE/swift-5.9.2-RELEASE-osx.pkg
```

Package.swift
```diff
@@ -1,10 +1,12 @@
-// swift-tools-version:5.7
+// swift-tools-version:5.10
```
0xTim, line 1: If we're bumping to 5.10 we should probably switch on strict concurrency checking

Sources/MacOSPlatform/MacOS.swift
```diff
@@ -0,0 +1,184 @@
...
+
+        SwiftlyCore.print("""
+            NOTE: On macOS it is possible that the shell will pick up the system Swift on the path
+            instead of the one that swiftly has installed for you. You can run the 'hash -r'
+            command to update the shell with the latest PATHs.
+
+                hash -r
+
+            """
+        )
```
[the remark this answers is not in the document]
0xTim: +1 to just doing it ourselves

Sources/MacOSPlatform/MacOS.swift
```diff
@@ -0,0 +1,184 @@
...
+            .appendingPathComponent("usr", isDirectory: true)
+            .appendingPathComponent("bin", isDirectory: true)
+
+        // Delete existing symlinks from previously in-use toolchain.
+        if let currentToolchain {
+            try self.unUse(currentToolchain: currentToolchain)
+        }
+
+        // Ensure swiftly doesn't overwrite any existing executables without getting confirmation first.
+        let swiftlyBinDirContents = try FileManager.default.contentsOfDirectory(atPath: self.swiftlyBinDir.path)
```
0xTim, line 76: I wonder if it's worth us migration to NIOFileSystem, since we already have NIO as a dependency. Aside from the nicer API, it would be one less thing to use when depending on new Foundation which should make for a smaller binary

install/swiftly-install.sh
cmcgee1024: > GNU `getopt` is possible to install on macOS, so we could attempt to detect it or require it, but I'm not sure if its use is widespread enough to warrant either of those.

I tried seeing if we could ask users to install a GNU getopt on macOS easily, but even the homebrew version doesn't support what is needed by the installer script. Moving the installer script to Swift should hopefully give us much better, and more user-friendly options.

install/swiftly-install.sh
cmcgee1024: Thanks, I'll update those locations too. This is all abstracted nicely in the Swift code through the Platform protocol, so hopefully we can start using that soon instead of this duplication.

cmcgee1024:
> I don't want to push people into a position where they need Xcode installed to run with Swift (in the future I do hope we can install the swift toolchain without requiring Xcode). But I'm wondering, if Xcode is installed or at least `xcode-select` is available we use that for selecting the toolchain instead of pointing PATH to a folder of symbolic links. This would avoid the issue of clashing with /usr/bin/swift and work more in-sync with how Xcode works. Although this might mean we can't implement local proxies so I guess could be a deadend.

If users are using Xcode then Swiftly can help them to get toolchains installed and Xcode can pick them up from there, along with the xcode-select mechanisms. Otherwise, Swiftly will install the pkg files available from swift.org and provide a toolchain selection mechanism via use, the config.json, and either the symlinks/proxies (in the future). There's no need for Xcode if the user doesn't want to use it.

Sources/MacOSPlatform/MacOS.swift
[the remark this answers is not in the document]
cmcgee1024: Hash is a shell builtin, so we don't really have the ability at this level to run it on behalf of the user. I'm open to better suggestions on where to put this message at this stage. Coming down the pipe with the swiftly self-installation this message gets folded into the init sub-command that only gets shown at install/upgrade time.

Sources/Swiftly/Config.swift
```diff
@@ -56,7 +56,12 @@ public struct Config: Codable, Equatable {
             error: \"\(error)\".
             To use swiftly, modify the configuration file to fix the issue or perform a clean installation.
             """
+            #if !os(macOS)
             throw Error(message: msg)
+            #else
+            let pd = PlatformDefinition.init(name: "xcode", nameFull: "osx", namePretty: "macOS", architecture: nil)
+            return Config.init(inUse: nil, installedToolchains: [], platform: pd)
+            #endif
```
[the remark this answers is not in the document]
cmcgee1024: The motivation was to have Swiftly be capable of being run without the installer script in dev, or even through other types of installers (e.g. homebrew). The test strategy for macOS is manual at the moment, and there's no release for it yet for the install script to download. There needs to be a way to get a swiftly binary onto a Mac and run through tests.

MacOS doesn't require all of the auto-detection of platform like Linux. This all becomes more uniform in the future with Swiftly self-install through the init subcommand.

Package.swift
cmcgee1024: Thanks, I'm definitely interested in enabling better checks. How does one turn on the strict concurrency checking?

Sources/MacOSPlatform/MacOS.swift
cmcgee1024: The installer won't be running in certain cases, such as when the swiftly binary is being tested in dev, or if the swiftly binary is distributed through other means, such as packages. On macOS the testing is done manually at the moment. With swirly self-install and the init subcommand this gets centralized more.

DESIGN.md
cmcgee1024: I've installed some very old releases of the toolchain, such as around 3.x, in my testing with macOS. I'll remove some of these notes as the picture is different now.

Sources/MacOSPlatform/MacOS.swift
cmcgee1024: I can raise this as a separate issue. I believe that there are Foundation dependencies all over the code. Pulling all of that out would be a significant sweep.

I'm not familiar with NIOFileSystem. Something that I really like about FileManager API is that it appears to be very mockable so that we could refactor the tests to operate on a full mock of the filesystem. Is that capability available with NIOFS?

Sources/MacOSPlatform/MacOS.swift
```diff
@@ -0,0 +1,184 @@
...
+    public func getTempFilePath() -> URL {
+        FileManager.default.temporaryDirectory.appendingPathComponent("swiftly-\(UUID()).pkg")
+    }
+
+    public func verifySignature(httpClient: SwiftlyHTTPClient, archiveDownloadURL: URL, archive: URL) async throws {
+        // No signature verification is required on macOS since the pkg files have their own signing
+        //  mechanism and the swift.org downloadables are trusted by stock macOS installations.
+    }
+
+    private func runProgram(_ args: String..., quiet: Bool = false) throws {
```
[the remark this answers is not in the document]
cmcgee1024: I'm a little hesitant to do that since I expect that the `/usr/bin/env` path trick might not work with Windows, so only 2/3 platforms can use this as-is. I know that duplication can be painful at times, but so can leaky abstractions.

install/swiftly-install.sh
cmcgee1024: I've switched this over to uname.

DESIGN.md
cmcgee1024: I have fixed the references.

Sources/SwiftlyCore/Platform.swift
cmcgee1024: This has been replaced with a Platform abstraction for both the toolchain and bin directories.

Sources/SwiftlyCore/ToolchainVerison.swift
```diff
@@ -175,6 +175,24 @@ public enum ToolchainVersion {
             }
         }
     }
+
+    public var identifier: String {
```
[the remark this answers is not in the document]
cmcgee1024: It appears so. I've changed the Install.swift in the latest version to use the identifier for assembling the end part of the url path.

Sources/MacOSPlatform/MacOS.swift
```diff
@@ -0,0 +1,184 @@
...
+            throw Error(message: "could not decode plist at \(infoPlist)")
+        }
+
+        try FileManager.default.removeItem(at: toolchainDir)
+
+        let homedir = ProcessInfo.processInfo.environment["HOME"]!
+        try runProgram("pkgutil", "--volume", homedir, "--forget", pkgInfo.CFBundleIdentifier)
+    }
+
+    public func use(_ toolchain: ToolchainVersion, currentToolchain: ToolchainVersion?) throws -> Bool {
```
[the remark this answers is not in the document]
cmcgee1024: There's enough difference that I'm not sure what is a good abstraction here. Also, I expect that when we go to a proxy model these basically go away because there's no symlinks to retarget.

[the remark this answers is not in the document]
patrickfreed: The only difference appears to be the toolchain directory and the warning about rehashing:
```diff
--- /home/patrick/swiftly/Sources/LinuxPlatform/current.txt	2024-06-17 22:51:14.487659136 -0400
+++ /home/patrick/swiftly/macos.txt	2024-06-17 23:01:42.160478308 -0400
@@ -1,7 +1,6 @@
-
     public func use(_ toolchain: ToolchainVersion, currentToolchain: ToolchainVersion?) throws -> Bool {
         let toolchainBinURL = self.swiftlyToolchainsDir
-            .appendingPathComponent(toolchain.name, isDirectory: true)
+            .appendingPathComponent(toolchain.identifier + ".xctoolchain", isDirectory: true)
             .appendingPathComponent("usr", isDirectory: true)
             .appendingPathComponent("bin", isDirectory: true)
 
@@ -42,12 +41,22 @@
             )
         }
 
+        SwiftlyCore.print("""
+            NOTE: On macOS it is possible that the shell will pick up the system Swift on the path
+            instead of the one that swiftly has installed for you. You can run the 'hash -r'
+            command to update the shell with the latest PATHs.
+
+                hash -r
+
+            """
+        )
+
         return true
     }
 
     public func unUse(currentToolchain: ToolchainVersion) throws {
         let currentToolchainBinURL = self.swiftlyToolchainsDir
-            .appendingPathComponent(currentToolchain.name, isDirectory: true)
+            .appendingPathComponent(currentToolchain.identifier + ".xctoolchain", isDirectory: true)
             .appendingPathComponent("usr", isDirectory: true)
             .appendingPathComponent("bin", isDirectory: true)
 
@@ -57,7 +66,7 @@
             }
 
             let url = self.swiftlyBinDir.appendingPathComponent(existingExecutable)
-            let vals = try url.resourceValues(forKeys: [.isSymbolicLinkKey])
+            let vals = try url.resourceValues(forKeys: [URLResourceKey.isSymbolicLinkKey])
 
             guard let islink = vals.isSymbolicLink, islink else {
                 throw Error(message: "Found executable not managed by swiftly in SWIFTLY_BIN_DIR: \(url.path)")


```
For the toolchain directory, we can add something like the following to the `Platform` protocol:
```swift
func getToolchainDirectoryName(for toolchain: ToolchainVersion) -> String
```
On Linux, this could return `toolchain.name`, and on macOS it can return `toolchain.identifier + ".xctoolchain"`.

Sources/MacOSPlatform/MacOS.swift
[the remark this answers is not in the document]
patrickfreed: This method gets called every time a different toolchain is used, so that includes even when just switching between already installed toolchains. In light of that, I'd say it'd be best to move it to `Install.swift`, and only print it after installing + using the first toolchain, only if there is already an existing Swift installation on the PATH. Since this is more of a bash thing, I don't think it is specific to macOS, so we can also drop that part of the message.

Sources/MacOSPlatform/MacOS.swift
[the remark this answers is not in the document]
patrickfreed: Yeah, I agree that once we start to introduce Windows support, we won't want it as a default implementation, but until then, I think it'd be best to cut down on the duplication though. As an alternative, this logic could be factored out into a separate `unixRunProgram` function that both implementations call. Either way is fine with me.

Sources/Swiftly/Config.swift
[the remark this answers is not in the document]
patrickfreed: The existing way to do this is to set the following environment variables:
- SWIFTLY_PLATFORM_NAME
- SWIFTLY_PLATFORM_NAME_FULL
- SWIFTLY_PLATFORM_NAME_PRETTY

You can see example values for these files in the various dockerfiles that were previously used for CI: 
https://github.com/swiftlang/swiftly/blob/ab38db009304334aebd3c661126eba8b2b1536b7/docker/docker-compose.2204.yaml#L14-L16 

I think it'd be best if we continued to rely on these environment variables for this, since changing this code might affect non-test usage of swiftly.

Sources/MacOSPlatform/MacOS.swift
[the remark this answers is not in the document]
adam-fowler: I think we can leave this to a later PR, if it is needed.

Package.swift
adam-fowler: In your target definition in Package.swift add
```swift
swiftSettings: [.enableExperimentalFeature("StrictConcurrency=complete")]
```

Sources/MacOSPlatform/MacOS.swift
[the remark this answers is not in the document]
cmcgee1024: An issue with zsh on macOS with the symlinks is that any change to the symlinks might invalidate zsh's internal path cache and it might start picking up the system's /usr/bin/swift instead of the symlink. I think that the message might be relevant any time a toolchain is used and the symlinks change. I hope that this becomes better with the proxies since the symlinks are stable until a swiftly init/self-update that can potentially change the symlinks, which is much less frequent.

Meanwhile, I think the message is annoying, but relevant at this point in time as someone might hit this in its current state.

Package.swift
cmcgee1024: Thanks, I tried enabling this and there were a non-trivial number of warnings (errors in Swift 6). I've raised an issue #124  to address the concurrency problems and enable the check.

Sources/MacOSPlatform/MacOS.swift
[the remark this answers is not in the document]
cmcgee1024: I've added a Linux/macOS only runProgram to the protocol, guarded with `#if` in case another platform, like Windows, tries to use it.

Sources/MacOSPlatform/MacOS.swift
0xTim: To be clear I'm not suggesting we pull out all of Foundation, but minimising it to just `FoundationEssentials` etc could help reduce the size (especially if we don't need ICU).

Which specific parts would you use for mocking (not getting into a discussion about that word 🤣 )? NIOFileSystem has a _similar_ API in that you have `FileSystem.shared` you perform the operations on, but the API is much more modern and built with Concurrency in mind (may also help solve some of the issues in #124 )

FranzBusch: In general, I agree that NIOFileSystem is more appropriate here since it uses async code to do the work. However, I would defer that for this PR since it is something we can do separately. WDYT @0xTim ?

adam-fowler: Yes let's defer this to another PR. I've added an issue #125

0xTim: Yeah absolutely shouldn't be a blocker for this PR

cmcgee1024:
Thanks everyone. @adam-fowler @patrickfreed is there anything left that's blocking this from being merged?

adam-fowler:
> Thanks everyone. @adam-fowler @patrickfreed is there anything left that's blocking this from being merged?

Do you think you could get the tests running? Although that might be quite a large undertaking. @shahmishal do you think we could get macOS CI up and running?

cmcgee1024:
> Do you think you could get the tests running? Although that might be quite a large undertaking. @shahmishal do you think we could get macOS CI up and running?

The plan for macOS right now is just ad-hoc testing, since much of the core swiftly logic remains the same here and is covered by the install and xctest cases. I believe that those are all still functional with this patch, but it will be good to get the CI hooked back up so that you all don't have to take my word for it.

I would like to revisit testing in the near future to see if we can make the xctests work in full mocked isolation, rework the installation tests to use something Swift-ish instead of directly depending on docker/docker-compose.

adam-fowler:

adam-fowler:
@swift-server-bot add to whitelist

adam-fowler:
@shahmishal I don't seem to be able to trigger builds

adam-fowler:
@cmcgee1024 so I tried to use this and kept getting a `NIOTooManyBytes` error thrown. By the time I worked out where the error was coming from I started getting `API rate limit exceeded ...` errors so couldn't test anymore.

Sources/MacOSPlatform/MacOS.swift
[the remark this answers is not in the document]
patrickfreed: Have we confirmed that this happens even when just switching from one toolchain to another? It would be pretty surprising behavior if invalidating the cache caused an item later in the PATH search tree to be picked up, since while the destination of the symlink changes, the link itself will still be in the same place. Also, do we know what the macOS specific part of this is? Since it’s zsh related, I would have imagined wed see similar behavior on Linux.

Regardless, I really do think switching to a toolchain is far too common of an operation to be printing this message each time, and it runs the risk of obscuring more important output. I think if we want to be safe, printing it the first time a user installs a toolchain and including it in the ‘—help’ output is sufficient.

patrickfreed (commented):
My only remaining asks are reducing the frequency of the PATH hash warning message and removing the macOS specific configuration default when the file is missing.

patrickfreed:
FYI you can bypass the API rate limit error by passing a permissionless GitHub token to your swiftly command invocation.

cmcgee1024:
> @cmcgee1024 so I tried to use this and kept getting a `NIOTooManyBytes` error thrown. By the time I worked out where the error was coming from I started getting `API rate limit exceeded ...` errors so couldn't test anymore.

Is this happening when you try to run the tests from macOS, or Linux? For macOS, I don't think that they're going to work in their current form. In either case, I'm not sure how these changes in this PR cause NIOTooManyBytes, so I'd be curious if you get the same thing with the latest in main.

The rate limits can be overcome using a GitHub personal token from your account and no permissions at all since the releases API of a public project is public. You can give it to swiftly using an environment variable like this:

```
SWIFTLY_GITHUB_TOKEN=github_pat_ABCDEFG...
```

I hope that we can mock these much better in the future so that they don't hit rate limits, susceptible for external conditions, filesystem state. There'll be a few integration and end-to-end tests to check for smoke, but I expect these to much fewer in number, so easier to retry when they fail.

Sources/MacOSPlatform/MacOS.swift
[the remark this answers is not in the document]
cmcgee1024: I think what makes macOS so special in this respect is that its default shell is zsh, and many/most installs will have a conflicting swift binary on the user's PATH, with a dialog that can be really confusing for people using swiftly. It's this combination that I feel warrants to have a relatively loud message in this state of this commit for the sake of usability.

cmcgee1024:
> My only remaining asks are reducing the frequency of the PATH hash warning message and removing the macOS specific configuration default when the file is missing.

The type of testing that I'm doing on macOS at the moment is ad-hoc. It involves copying over the binary onto a fresh system, and running smoke tests. This auto-configuration is really helpful for this until we have the init subcommand in an upcoming patch. Can we keep this in place for the sake of testing so that I don't have to create a custom version of the installer shell script?

cmcgee1024:
I've reproduced the NIOTooManyBytesError on Ubuntu 22.04 Linux with the swift docker image, and it's also happening on main. I've opened a new issue #128  to track this down.

adam-fowler:
> The plan for macOS right now is just ad-hoc testing, since much of the core swiftly logic remains the same here and is covered by the install and xctest cases. I believe that those are all still functional with this patch, but it will be good to get the CI hooked back up so that you all don't have to take my word for it.

Do you mean, you believe the XCTests are working for macOS? Because currently most of them are failing on macOS.

cmcgee1024:
@adam-fowler I've just confirmed that the xctests are working in Linux on this branch as well as main as long as the GitHub token is provided. It appears that without a token the GH API's sometimes give inaccurate content lengths under load. I expect that the tests do not work on macOS, and so the plan is ad-hoc testing on there.

shahmishal:

shahmishal:
@swift-ci test macOS

cmcgee1024:

shahmishal:

shahmishal:
@swift-ci test macOS

patrickfreed:
>The type of testing that I'm doing on macOS at the moment is ad-hoc. It involves copying over the binary onto a fresh system, and running smoke tests. This auto-configuration is really helpful for this until we have the init subcommand in an upcoming patch. Can we keep this in place for the sake of testing so that I don't have to create a custom version of the installer shell script?

Per my [comment](https://github.com/swiftlang/swiftly/pull/121#discussion_r1643689047) above, there are a few environment variables that swiftly reads from for this purpose, so there's no need for a custom version of the installer script. Something as simple as the following would work just fine:
```bash
export SWIFTLY_PLATFORM_NAME=xcode
export SWIFTLY_PLATFORM_NAME_FULL=osx
export SWIFTLY_PLATFORM_NAME_PRETTY=macOS
swift test
```
Or you could also turn it into a one-liner if you drop the `export`s.

If we'd like to change how swiftly behaves when there isn't a config present, I think that's fine, but we should do that in a separate PR rather than as a side effect of introducing macOS support. We'd also want to be sure to be consistent across platforms for something like that.

Sources/MacOSPlatform/MacOS.swift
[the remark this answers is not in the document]
patrickfreed: If we plan to remove this message once we switch to using proxies, then the discussion is sort of moot anyways, so we can just leave it in for now. If we don't plan on removing it, then I'd like to discuss this again at that time, since I feel pretty strongly that this message is too noisy to be printed outside of installation of swiftly and/or installation of the first swiftly-managed toolchain (we already do the former actually, see [here](https://github.com/swiftlang/swiftly/blob/364b02bb2904cff62a604bfa4a0ef0d1d9164cd5/install/swiftly-install.sh#L672)).

cmcgee1024:

cmcgee1024:
@swift-ci test macOS

shahmishal:
@swift-ci test macOS

shahmishal:
@swift-ci test macOS

cmcgee1024:

cmcgee1024:
@swift-ci test macOS

adam-fowler (approved):
All looks good now

adam-fowler:
@swift-ci test macOS

patrickfreed:

patrickfreed:

patrickfreed:
@swift-ci test macOS

patrickfreed:
Hmm, not seeing any macOS tasks get spawned here.

adam-fowler:
> Hmm, not seeing any macOS tasks get spawned here.

@shahmishal It doesn't seem like anyone can kick off the macOS testing

shahmishal:
Can you try again? I think I fixed it now

adam-fowler:
@swift-ci test macOS

adam-fowler:
> Can you try again? I think I fixed it now

Doesn't look like it, am I an admin though? I seem to remember you said you only set Patrick as the admin

patrickfreed:
@swift-ci test macOS

patrickfreed:
Doesn't seem like it worked for me either.

shahmishal:
@swift-ci test macOS

cmcgee1024:
The three Linux pull request validations appear to be a timing issue. I'll update with a fix for them.

cmcgee1024:

shahmishal:

adam-fowler:
@cmcgee1024 Looks like we have quite a few macOS fails here

cmcgee1024:
> @cmcgee1024 Looks like we have quite a few macOS fails here

Yes, while the tests work in dev, there are some differences with the CI. There's two flavours of failures. The first is that there are restrictions affecting the installer for the build user. I'm currently looking at how to mock this out so that CI can pass.

The second flavour is related to network access. Something appears to be blocking some or all of the network access. While I think that most network access in the tests should be mocked in the longer term, there will still need to be some tests (probably far fewer in number) that exercise the various API's that swiftly uses to check for sanity.

cmcgee1024:
@swift-ci test macOS

patrickfreed:
Bumping https://github.com/swiftlang/swiftly/pull/121#issuecomment-2182924205 just in case it got lost in the testing comments. It's my only outstanding comment besides addressing the test failures.

As a side note, GitHub PRs/reviews don't handle force pushes all that well, so it's usually better to just push more commits to the PR's branch. That way, links to prior conversations will remain valid (as an example, [this link](https://github.com/swiftlang/swiftly/pull/121/files/58e9dee3245d7a7cd542be9c57af80d8ec99cc2a#r1641454583) to one of my comments no longer works because the commit it was associated with was overwritten by a force push), and reviewers can more easily look back at the incremental changes being made. When merging the PR, you can just squash all the intermediate commits into a single one via the GitHub UI anyways (there will be a "squash and merge" option).

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
@swift-ci test macOS

cmcgee1024:

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
@swift-ci test macOS

cmcgee1024:

shahmishal:
@swift-ci test macOS

shahmishal:

adam-fowler:
Hi @cmcgee1024 
Coming back to this, looks like you have one issue with `InstallTests.testInstallUsesFirstToolchain` on some of the linux builds, any thoughts on this?

cmcgee1024:
@adam-fowler it appears that the signature file was empty or corrupted on three out of the four linux validation checks. They're all running the same tests. I wonder if this is some kind of transient failure at the time the tests were run.

cmcgee1024:

Tests/SwiftlyTests/SwiftlyTests.swift
```diff
@@ -461,4 +535,67 @@ public struct MockToolchainDownloader: HTTPRequestExecutor {
 
         return try Data(contentsOf: archive)
     }
+
+#elseif os(macOS)
```
patrickfreed, line 539: The first two halves of this branch are identical, can we combine them and only have the OS-specific parts inside the branches?

Tests/SwiftlyTests/SwiftlyTests.swift
```diff
@@ -6,11 +6,71 @@ import NIO
...
 class SwiftlyTests: XCTestCase {
+    private static var requestExecutor: ProxyHTTPRequestExecutorImpl?
+
+    override class func setUp() {
+        if Self.requestExecutor == nil {
+            Self.requestExecutor = ProxyHTTPRequestExecutorImpl()
+            Install.httpClient = SwiftlyHTTPClient(executor: Self.requestExecutor)
+            SelfUpdate.httpClient = SwiftlyHTTPClient(executor: Self.requestExecutor)
+        }
+    }
```
patrickfreed, line 65: I think we'll need to unconditionally reset all the httpClients to use the proxy executor, since many tests will replace them with mocked http clients. To make this easier, it would probably be preferable to have one single global singleton HTTP client defined in SwiftlyCore or something.

patrickfreed (commented):
We'll also want to make sure the swiftly installation tests work on macOS before merging.

shahmishal:
@swift-server-bot add to allowlist

cmcgee1024:

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
@patrickfreed I've refactored the way that the http client is shared and mocked and all of the tests are passing on both Linux and macOS. I hope that this is ready to squash and merge at this point. Otherwise, is there anything that still needs to be addressed?

cmcgee1024:
I'm merging this PR since the tests are now passing and the discussions points seem to have converged. Please raise issues for anything that remains. Thanks everyone!

patrickfreed (approved):
Sorry for the delay, the only outstanding comment I have is [this](https://github.com/swiftlang/swiftly/pull/121/files#r1664172553) about deduplicating some code, but otherwise looks good to me. Feel free to defer that to a follow on if you'd like. Excited to see swiftly running on macOS!

patrickfreed:
Oh I also think it would be good to verify the installation tests still work as well

patrickfreed:

PR #127, opened 2024-06-19
Merge installation logic into swiftly as an init subcommand

cmcgee1024:
Instead of relying on a separate shell script with certain issues swiftly can perform its own user initialization with many of the features of the script, including prompts, flags, and options. It also prepares swiftly for possible delivery through system package managers.

Provide a design for the new installation system in the DESIGN.md.

Add an implementation, tests and documentation of an init subcommand that is capable of finishing the installation of swiftly (after the download step) into the user's environment, including swiftly home, toolchains directory, the swiftly binary (if it's user downloaded), and symbolic links.

Contribute documentation describing how to install swiftly and toolchains automatically in the context of a build/CI system.


DESIGN.md
```diff
@@ -19,19 +19,35 @@ This document contains the high level design of swiftly. Not all features have b
...
-- CentOS 8
-- Amazon Linux 2
+We'll need an initialization mode which, detects information about the OS and distribution in the case of Linux. The initialization mode is also responsible for setting up the directory structure for the toolchains (if necessary), setting up the shell environment for the user, and determining any operating system level dependencies that are required, but missing. Swiftly can perform these tasks itself with the capabilities provided by the Swift language and libraries, such as rich argument parsing, and launching system processes provided that the binary is somehow delivered to the user.
+
+The delivery of the binary onto the user's system can be accomplished in a number of ways, such as direct download from a trusted website with some guidance for the user to pick the right one for their OS and architecture. There might also be a copy/paste of a simple shell command-line for some more automation. Swiftly might be provided as a system-level package (e.g. homebrew, apt, yum, macOS pkg, etc). In some cases delivery might come from manual compilation from this git repository and the "swift build" command, useful for swiftly development and testing.
```
adam-fowler, line 26: Delivery of the binary is where this falls down to some degree (at least on Linux). Previously this was all done automatically by the script. macOS has pkg which avoids this issue, but unless we are using some package management system it will up to the user to copy the binary into a folder in their path which isn't an ideal situation. Or are you expecting `swiftly-init` to do this for the user?

cmcgee1024: All of the path/environment management, creating the swiftly configuration, detection of missing required system packages, and most of what the shell script does today would be incorporated into swiftly's new init subcommand in this design. It's up to the delivery mechanisms to help the user to get to that entry point. I think that with some of the conventions described here those mechanisms have some mechanisms to provide optimal user experiences. For example, I think that swift.org can provide a copy-able command-line that would be equivalent to what is there today, but a little more complex.

0xTim: I think it would be good to breakdown the assumed different responsibilities of the different delivery methods to ensure that we're covering all bases and don't have any missing features or regressions in behaviours. This would help us hash out what needs to be done for different platforms, what needs to be be done for security etc

DESIGN.md
```diff
@@ -19,19 +19,35 @@ This document contains the high level design of swiftly. Not all features have b
...
+
+```
+swiftly init
+```
+
+Since an end user may not be aware that swiftly has an initialiation step, its possible that they will invoke it in an uninitialized state on their account. If this happens then swiftly validates its own installation every time it is invoked and helps the user to discover the init subcommand. A system package, or website, can use an accelerator to help guide the user to the init subcommand by naming the swiftly binary as `swiftly-init*`. Swiftly discovers that it is being invoked in this way and automatically runs the initialization right away. If the system package installs swiftly-init in a location that's on the user's default path, tab completion of `swiftly` will usually complete to the full name and the user is directed to run the initialization by default. However, if swiftly is already on the user's path then the completion will be unintrusive.
```
patrickfreed, line 36: I think having a separate `swiftly-init` binary overcomplicates things here. I think simply printing a message indicating the user should run `swiftly init` if they try to run a command without having done so should be sufficient.

DESIGN.md
```diff
@@ -19,19 +19,35 @@ This document contains the high level design of swiftly. Not all features have b
...
+
+```
+swiftly-init-0.4.0 # Runs 'swiftly init' subcommand to initialize swiftly
+```
+
+### Swiftly upgrades
+
+It is possible that initialization will detect an existing swiftly installation at an older version, which then performs an upgrade procedure. If the existing installation is newer (using the `--version flag`) then the initialization is aborted. This gives the user the opportunity to upgrade to a newer version of swiftly by manually downloading it from a trusted website or ad-hoc test a pre-release.
```
patrickfreed, line 44: I would imagine that the case we should be solving for here is not that there are multiple binaries on the system, but rather that a newer swiftly binary was dropped in when an older configuration existed. In that case, we'd detect this via some marker in the `config.json`, rather than by invoking `--version`. I don't think we'll want swiftly combing the PATH for other versions of swiftly really, and we can just rely on the contents of `config.json`.

Sources/SwiftlyCore/SwiftlyCore.swift
```diff
@@ -1,6 +1,6 @@
 import Foundation
```
patrickfreed, line 1: These SwiftlyVersion changes seem unrelated to the proposed changes to the design, can we omit them from this PR? We can open another PR for them.

DESIGN.md
```diff
@@ -19,19 +19,35 @@ This document contains the high level design of swiftly. Not all features have b
...
-- Ubuntu 16.04
-- Ubuntu 18.04
-- Ubuntu 20.04
-- CentOS 7
-- CentOS 8
-- Amazon Linux 2
+We'll need an initialization mode which, detects information about the OS and distribution in the case of Linux. The initialization mode is also responsible for setting up the directory structure for the toolchains (if necessary), setting up the shell environment for the user, and determining any operating system level dependencies that are required, but missing. Swiftly can perform these tasks itself with the capabilities provided by the Swift language and libraries, such as rich argument parsing, and launching system processes provided that the binary is somehow delivered to the user.
```
patrickfreed, line 24: What does "mode" refer to here? A subcommand?

patrickfreed, line 24: I think we discussed in the meeting that it would be acceptable for `swiftly init` to install these packages, but just not by default.

Sources/SwiftlyCore/SwiftlyCore.swift
cmcgee1024: It is meant to be foundational work for the upgrade detection scenarios in the design. I can raise a separate PR if you feel it would be better to get it in right away.

DESIGN.md
cmcgee1024: Thanks, I'll try a simpler design here that doesn't have the swiftly-init accelerator mode. The inspiration for this was the rustup-init homebrew formula, but I see now that rustup deb and rpm don't use the same approach.

DESIGN.md
cmcgee1024: I'll restructure this paragraph to add some more clarity.

cmcgee1024:
@adam-fowler @patrickfreed how does this new revision of the design look? I've enumerated the expected delivery methods, simplified things, and fixed up some of the wording.

DESIGN.md
```diff
@@ -19,19 +19,32 @@ This document contains the high level design of swiftly. Not all features have b
...
+```
 
-We'll need a bootstrapping script which detects information about the OS and downloads the correct pre-built swiftly executable. We can use [rustup-init.sh](https://github.com/rust-lang/rustup/blob/master/rustup-init.sh) as a general guide on implementing such a script, though it is more complicated and supports far more systems than I think we need to. At least for the initial release, I think we'll only need to support the platforms listed on [Swift.org - Getting Started](http://swift.org/getting-started), namely:
-- Ubuntu 16.04
-- Ubuntu 18.04
-- Ubuntu 20.04
-- CentOS 7
-- CentOS 8
-- Amazon Linux 2
+Initialization also covers creating symbolic links for various toolchain commands (e.g. swift, swiftc, clang, clang++, etc.) so that swiftly can configure the targets. These need to be installed into a user controlled directory so that swiftly can operate on the without super user permissions. Also, the swiftly binary itself is copied into this location if it is not run from a system location, managed by a system package manager.
```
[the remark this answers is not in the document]
cmcgee1024: Thanks, I've reworked this paragraph to notify the user about the case where swiftly is copied into the swiftly bin directory so that they are aware.

0xTim (approved):

patrickfreed (commented):
sorry for the late review, mostly looks good just have a few minor suggestions

DESIGN.md
```diff
@@ -19,19 +19,32 @@ This document contains the high level design of swiftly. Not all features have b
...
+* Direct download from a trusted website with guidance on the correct binary for the user's platform to download and then how move on to initialization
+* System-level package (e.g. homebrew, pkg, apt-get, rpm) that downloads and places the swiftly binary in a system location outside of the user's home directory, often runnable from the user's path
+* Manual compilation of the swiftly binary from this git repository (e.g. from a development environment)
+
+We'll need an initialization phase, which detects information about the OS and distribution in the case of Linux. The initialization mode is also responsible for setting up the directory structure for the toolchains (if necessary), setting up the shell environment for the user, and determining any operating system level dependencies that are required, but missing. Swiftly has its own configuration stored in the form of a `config.json` file, which will be created as part of initialization. None of the delivery methods can perform all of these steps on their own. System package managers don't normally update all users' profile or update the user's home directory structure directly.
```
patrickfreed, line 31: I think it's worth mentioning the creation of an `env.sh` containing environment variable definitions (e.g. SWIFTLY_HOME_DIR, SWIFTLY_BIN_DIR, and PATH) and updating the `.profile`/`.bash_profile`/`.zshrc` to source it.

e.g. this is what `swiftly-install.sh` creates by default today
```bash
export SWIFTLY_HOME_DIR="$HOME/.local/share/swiftly"
export SWIFTLY_BIN_DIR="$HOME/.local/bin"
if [[ ":$PATH:" != *":$SWIFTLY_BIN_DIR:"* ]]; then
   export PATH="$SWIFTLY_BIN_DIR:$PATH"
fi
```

DESIGN.md
```diff
@@ -19,19 +19,32 @@ This document contains the high level design of swiftly. Not all features have b
...
-- Ubuntu 16.04
-- Ubuntu 18.04
-- Ubuntu 20.04
-- CentOS 7
-- CentOS 8
-- Amazon Linux 2
+The swiftly binary itself is moved (or copied as a fallback) into the SWIFTLY_BIN_DIR location (or platform default) if it is not run from a system location where it is managed by a system package manager. If the binary could not be moved then the user is notified that they can remove the original.
 
-Once it has detected which platform the user is running, the script will then create `$HOME/.local/share/swiftly` (or a different path, if the user provides one. For an initial MVP, I think we can always install there). It'll also create `$HOME/.local/bin` if needed, download the prebuilt swiftly executable appropriate for the platform, and drop it in there.
+## Swiftly update
```
patrickfreed, line 41: can we rename this `Updating swiftly` to disambiguate from the `swiftly update` command?

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
@swift-ci test macOS

patrickfreed:
Can we separate out the code and docs changes into another PR? The standing approvals for this PR were made only for the design changes.

patrickfreed:
The design changes look good to me by the way, sorry I thought I had approved that part a while back.

cmcgee1024:
> Can we separate out the code and docs changes into another PR? The standing approvals for this PR were made only for the design changes.

I was thinking of keeping the design, and artifacts together in one commit, so that there's no point in history where the design disagrees with the implementation. I can reset all of the reviews here so that approvals reflect the current scope.

patrickfreed:
Yeah if that workflow is better for you that's totally fine. I just wanted to make sure reviewers were made aware that the PR has grown in scope since their approval, so resetting that status sounds good to me.

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
@adam-fowler @patrickfreed this should be ready for review now. The macOS tests all pass, but there's some kind of CI system problem that throws errors at the end. Thank you.

Sources/LinuxPlatform/Linux.swift
```diff
@@ -227,5 +414,196 @@ public struct Linux: Platform {
...
+                versionID = String(info.dropFirst("VERSION_ID=".count)).replacingOccurrences(of: "\"", with: "")
+            } else if info.hasPrefix("UBUNTU_CODENAME=") {
+                ubuntuCodeName = String(info.dropFirst("UBUNTU_CODENAME=".count)).replacingOccurrences(of: "\"", with: "")
+            } else if info.hasPrefix("PRETTY_NAME=") {
+                platformPretty = String(info.dropFirst("PRETTY_NAME=".count)).replacingOccurrences(of: "\"", with: "")
+            }
+        }
+
+        guard let id = id, let idlike = idlike else {
+            let message = "Unable to find release information from file \(releaseFile)"
```
adam-fowler, line 533:
```suggestion
            let message = "Unable to find OS release information from file \(releaseFile)"
```

Sources/LinuxPlatform/Linux.swift
```diff
@@ -227,5 +414,196 @@ public struct Linux: Platform {
...
+                throw Error(message: message)
+            } else {
+                print(message)
+            }
+            return self.manualSelectPlatform(platformPretty)
+        }
+
+        let data = FileManager.default.contents(atPath: releaseFile)
+        guard let data = data else {
+            let message = "Unable to read release information from file \(releaseFile)"
```
adam-fowler, line 495:
```suggestion
            let message = "Unable to read OS release information from file \(releaseFile)"
```

Sources/Swiftly/Init.swift
```diff
@@ -0,0 +1,206 @@
...
+            SwiftlyCore.print("""
+            Swiftly will be installed into the following locations:
+
+            \(Swiftly.currentPlatform.swiftlyHomeDir.path) - Data and configuration files directory including toolchains
+            \(Swiftly.currentPlatform.swiftlyBinDir.path) - Executables installation directory
+
+            Note that the locations can be changed with SWIFTLY_HOME and SWIFTLY_BIN environment variables and run
+            this again.
+
+            Proceed with the installation?
```
adam-fowler, line 50: Can we change this to a [Y/n]?

README.md
```diff
@@ -11,6 +11,11 @@ To download swiftly and install Swift, run the following in your terminal, then
 curl -L https://swiftlang.github.io/swiftly/swiftly-install.sh | bash
```
adam-fowler, line 11: We either need to edit this script to download the correct version of swiftly (Linux or mac) and run `swiftly init` or get rid of this line in the docs.

Documentation/SwiftlyDocs.docc/automated-install.md
```diff
@@ -0,0 +1,66 @@
...
+```
+./swiftly init --assume-yes
+```
+
+Swiftly is installed, but the current shell may not yet be updated with the new environment variables, such as the PATH. The init command prints instructions on how to update the current shell environment without opening a new shell.
+
+```
+To begin using installed swiftly from your current shell, first run the following command:
+
+. "/root/.local/share/swiftly/env.sh"
```
adam-fowler, line 26: Need to make it clear this is for Linux, maybe provide macOS version as well

Documentation/SwiftlyDocs.docc/getting-started.md
```diff
@@ -6,6 +6,12 @@ To download swiftly and install Swift, run the following in your terminal, then
 curl -L https://swiftlang.github.io/swiftly/swiftly-install.sh | bash
```
adam-fowler, line 6: We either need to edit this script to download the correct version of swiftly (Linux or mac) and run swiftly init or get rid of this line in the docs.

adam-fowler (changes requested):
Some minor change requests, but in general looks good.

Sources/Swiftly/Install.swift
```diff
@@ -59,36 +59,64 @@ struct Install: SwiftlyCommand {
...
         var config = try Config.load()
-        try await Self.execute(
+        let postInstallScript = try await Self.execute(
             version: toolchainVersion,
             &config,
             useInstalledToolchain: self.use,
             verifySignature: self.verify
         )
+
+        if let postInstallScript = postInstallScript {
```
adam-fowler, line 89: Is it ok to install the dependencies after the swift toolchain has been installed?

Sources/LinuxPlatform/Linux.swift
```diff
@@ -41,48 +41,227 @@ public struct Linux: Platform {
...
+                FileManager.default.createFile(atPath: tmpFile.path, contents: nil, attributes: [.posixPermissions: 0600])
+                defer {
+                    try? FileManager.default.removeItem(at: tmpFile)
+                }
+
+                guard let url = URL(string: "https://swift.org/keys/all-keys.asc") else {
+                    throw Error(message: "malformed URL to the swift gpg keys")
+                }
+
+                try await httpClient.downloadFile(url: url, to: tmpFile)
```
adam-fowler, line 215: I've found inside a vscode devcontainer `httpClient.downloadFile` fails to write to the tmpFile

Sources/LinuxPlatform/Linux.swift
```diff
@@ -41,48 +41,227 @@ public struct Linux: Platform {
...
+                "gpg",
+                "--list-keys",
+                "swift-infrastructure@forums.swift.org",
+                "swift-infrastructure@swift.org",
+                quiet: true
+            )) != nil
+            if !foundKeys {
+                // Import the swift keys if they aren't here already
+                let tmpFile = self.getTempFilePath()
+                FileManager.default.createFile(atPath: tmpFile.path, contents: nil, attributes: [.posixPermissions: 0600])
```
adam-fowler, line 206:
```suggestion
                FileManager.default.createFile(atPath: tmpFile.path, contents: nil, attributes: [.posixPermissions: 0o600])
```

Sources/LinuxPlatform/Linux.swift
adam-fowler: Worked out why octal literals are written `0o600`.

adam-fowler (changes requested):
I've added a suggested fix for an issue on Linux when downloading gpg keys

Sources/Swiftly/Init.swift
```diff
@@ -0,0 +1,206 @@
...
+
+    /// Initialize the installation of swiftly.
+    internal static func execute(assumeYes: Bool, noModifyProfile: Bool, overwrite: Bool, platform: String?) async throws {
+        try Swiftly.currentPlatform.verifySwiftlySystemPrerequisites()
+
+        let config = try? Config.load()
+
+        if let config = config, !overwrite && config.version != SwiftlyCore.version {
+            // We don't support downgrades, and we don't yet support upgrades
+            throw Error(message: "An existing swiftly installation was detected. You can try overwriting it.")
```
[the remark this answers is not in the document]
adam-fowler: Ah I understand, you need to add `--overwrite`. Maybe make the error more explicit. Also when running with `--overwrite` on Linux it returns a different error 
```
Error: The operation could not be completed. A file with the same name already exists.
```

Sources/Swiftly/Init.swift
```diff
@@ -0,0 +1,206 @@
...
+        let systemManaged = try Swiftly.currentPlatform.isSystemManagedBinary(cmd)
+
+        // Don't move the binary if this is being invoked inside an xctest or is a system managed binary
+        if !cmd.hasSuffix("xctest") && !systemManaged {
+            SwiftlyCore.print("Moving swiftly into the installation directory...")
+            if swiftlyBin.fileExists() && !overwrite {
+                throw Error(message: "Swiftly binary already exists. You can try again with overwrite to replace it.")
+            }
+            do {
+                try FileManager.default.moveItem(at: URL(fileURLWithPath: cmd), to: swiftlyBin)
```
adam-fowler, line 122: `moveItem` fails if the item exists already. Also should we check that cmd and swiftlyBin aren't exactly the same

Sources/Swiftly/Init.swift
```diff
@@ -0,0 +1,206 @@
...
+            } catch {
+                try FileManager.default.copyItem(at: URL(fileURLWithPath: cmd), to: swiftlyBin)
+                SwiftlyCore.print("Swiftly has been coopied into the installation directory. You can remove '\(cmd)'. It is no longer needed.")
```
adam-fowler, line 125:
```suggestion
                SwiftlyCore.print("Swiftly has been copied into the installation directory. You can remove '\(cmd)'. It is no longer needed.")
```

README.md
cmcgee1024: This is tricky since this README is one of the places where people see how to install swiftly, but it's a few commits newer than the last release. Also, it's unclear if we are going to cut a release or not that still makes use of the shell script, which will still technically work.

Documentation/SwiftlyDocs.docc/getting-started.md
cmcgee1024: It's the same kind of problem with the readme. People might be reading this from the main branch of swiftpackageindex.com and want the latest release, which is using the shell script installer. We'll need to sweep all of the documentation when we release to switch over to the new method.

Sources/Swiftly/Install.swift
cmcgee1024: Yes, I believe so. Swiftly doesn't need to run anything inside the toolchain, it just expands the tarball or macOS pkg. We just need the user to be aware of the dependencies that they don't have that the toolchain needs so that they can take those steps before using it. It's also possible in the future that newer toolchains may have more/less dependencies, or maybe none at all.

Documentation/SwiftlyDocs.docc/getting-started.md
adam-fowler: Ok we should create a separate PR with documentation changes

Sources/Swiftly/Init.swift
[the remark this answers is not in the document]
cmcgee1024: I've made the message more explicit about the `--overwrite` flag since it's in the init subcommand and the error can directly reference the subcommand's flag.

Sources/Swiftly/Config.swift
```diff
@@ -49,12 +27,18 @@ public struct Config: Codable, Equatable {
...
+                config.version = try? SwiftlyVersion(parsing: "0.3.0")
+            }
+            return config
         } catch {
             let msg = """
             Could not load swiftly's configuration file at \(Swiftly.currentPlatform.swiftlyConfigFile.path) due to
             error: \"\(error)\".
-            To use swiftly, modify the configuration file to fix the issue or perform a clean installation.
+            To begin using swiftly you can install it by running 'swiftly init'.
             """
```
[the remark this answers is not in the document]
cmcgee1024: I've pulled out this inner error since it's likely to cause more confusion. The focus should be on the instruction to run the init.

Sources/LinuxPlatform/Linux.swift
cmcgee1024: Thanks, this is much clearer.

Sources/Swiftly/Init.swift
cmcgee1024: I've changed the prompt to a yes or no instead of numbered options.

Sources/Swiftly/Init.swift
cmcgee1024: Thanks, I had assumed that the move and copy would overwrite. I've made an explicit remove in the case of overwrite and there is an existing binary.

Sources/Swiftly/Install.swift
adam-fowler: I guess the install is just unzipping a file in the end, so yeah just ignore this

Sources/SwiftlyCore/Platform.swift
```diff
@@ -106,6 +143,60 @@ extension Platform {
...
+        process.arguments = [program] + args
+
+        let outPipe = Pipe()
+        process.standardInput = FileHandle.nullDevice
+        process.standardError = FileHandle.nullDevice
+        process.standardOutput = outPipe
+
+        try process.run()
+        // Attach this process to our process group so that Ctrl-C and other signals work
+        let pgid = tcgetpgrp(STDOUT_FILENO)
```
adam-fowler, line 159: Can we add the same process group support to `runProgram`

Documentation/SwiftlyDocs.docc/automated-install.md
MahdiBM: `env.sh` is not accurate, for `fish`, it'll be `env.fish`.

README.md
MahdiBM: > reduced to copy the correct version of swiftly off the internet.

Yeah that's probably a good idea ... i think the maximum installation complexity i'd expect is "choose macOS or linux". Even that can probably be avoided in the script. Hopefully no manual architecture detection "are you on arm64 or amd64?!". manually choosing the arch is not the end of the world, but it can be harder than it looks, to beginners.

Sources/Swiftly/Init.swift
```diff
@@ -0,0 +1,208 @@
...
+        let cmd = URL(fileURLWithPath: CommandLine.arguments[0])
+        let systemManaged = try Swiftly.currentPlatform.isSystemManagedBinary(cmd.path)
+
+        // Don't move the binary if it's already in the right place, this is being invoked inside an xctest, or it is a system managed binary
+        if cmd != swiftlyBin && !cmd.path.hasSuffix("xctest") && !systemManaged {
+            SwiftlyCore.print("Moving swiftly into the installation directory...")
+
+            if swiftlyBin.fileExists() {
+                if !overwrite {
+                    throw Error(message: "Swiftly binary already exists. You can try again with overwrite to replace it.")
```
adam-fowler, line 117:
```suggestion
                    throw Error(message: "Swiftly binary already exists. You can try again with `--overwrite` to replace it.")
```

Documentation/SwiftlyDocs.docc/automated-install.md
cmcgee1024: > `env.sh` is not accurate, for `fish`, it'll be `env.fish`.

Right, I'm hoping to make this an illustrative example of something that an integrator would look for from the swiftly output so that they can automate it. I'll see if I can make the description above highlight that this is just an example and can be different depending on the environment (e.g. Windows or Fish)

Sources/SwiftlyCore/Platform.swift
cmcgee1024: I've added this functionality to the runProgram too.

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
@adam-fowler @patrickfreed although the macOS platform check is failing, that is happening after the tests due to some kind of problem with the CI system itself. The tests are all reporting as successful there.

adam-fowler (approved):

Issue #134, opened 2024-07-16
macOS: `$HOME/Library/Application Support` space not escaped in `swiftly-install.sh`


BrianHenryIE:
I built the repo and replaced `DOWNLOAD_URL` in `swiftly-install.sh` with the build I made.

The UI displays `$HOME/Library/Application` where it should show `$HOME/Library/Application Support/swiftly`

```
% ./swiftly-install.sh 

This script will install swiftly, a Swift toolchain installer and manager.

Current installation options:

    Data and configuration files directory: $HOME/Library/Application
        Executables installation directory: $HOME/Library/Application
     Modify login config ($HOME/.zprofile): true
               Install system dependencies: true

Select one of the following:
1) Proceed with the installation (default)
2) Customize the installation
3) Cancel
> 1
Downloading swiftly from https://brianhenryie.s3.amazonaws.com/2024/swiftly...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 18.7M  100 18.7M    0     0  5497k      0  0:00:03  0:00:03 --:--:-- 5499k

swiftly has been successfully installed!

Once you log in again, swiftly should be accessible from your PATH.
To begin using swiftly from your current shell, first run the following command:

    . $HOME/Library/Application/env.sh

Then to install the latest version of Swift, run 'swiftly install latest'

Warning: existing installation of Swift detected at /usr/bin/swift
To ensure swiftly-installed toolchains can be found by the shell, uninstall any existing Swift installation(s).
To ensure the current shell can find swiftly-installed toolchains, you may also need to run 'hash -r'.
```

When I open a new Terminal shell

```
/Users/brianhenry/.zprofile:.:9: no such file or directory: /Users/brianhenry/Library/Application/env.sh
```

and
```
% swiftly

zsh: command not found: swiftly
```

The binary was correctly installed
```
% ls $HOME/Library/Application\ Support/swiftly/bin/       

swiftly
```

but the `env.sh` file seems to have escaping problems
```
cat $HOME/Library/Application\ Support/swiftly/env.sh 

export SWIFTLY_HOME_DIR="$HOME/Library/Application"
export SWIFTLY_BIN_DIR="$HOME/Library/Application"
if [[ ":$PATH:" != *":$SWIFTLY_BIN_DIR:"* ]]; then
   export PATH="$SWIFTLY_BIN_DIR:$PATH"
fi
```

[...]

PR #135, opened 2024-07-16
`swiftly-install.sh` Handle paths containing spaces

BrianHenryIE:
In #134, the installer does not work on MacOS because of the space in the `$HOME/Library/Application Support/swiftly` path.

The first problem was that bash function calls split arguments by spaces. Using `$@` instead of `$1` fixed this.

Then the path added to the `source` commands needed to be quoted.

I don't know what effect these changes will have when run on Linux.

### DIFF
--- a/install/swiftly-install.sh
+++ b/install/swiftly-install.sh
@@ -648,7 +648,7 @@ fi
 EOF
                )
         ENV_FILE="env.sh"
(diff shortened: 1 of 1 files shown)


cmcgee1024:
We're in the process of moving the installation logic out of the shell script and into swiftly itself, which should fix the spacing issue since the Swift built-in path handling is immune to this kind of problem in shell scripts. See #127 for more details.

adam-fowler:
Now that #127 has been merged we can probably close this

Issue #149, opened 2024-08-06
Use swift.org metadata instead of GitHub API's for toolchain selection


cmcgee1024:
There are some comments in the code about using swift.org metadata about toolchain versions in toolchain search and selection instead of the GitHub API's. This would reduce the site dependencies of swiftly and help to make it dependent on one source of truth. This issue exists to track what is only mentioned in the code.

adam-fowler:
That PR referenced doesn't include the new platforms eg ubuntu2024

Do you know if since then these platforms are included?

cmcgee1024:
I'm not sure about the PR that added it, but ubuntu2204 seems to be working:

```
curl -L swift.org/api/v1/install/dev/6.0/ubuntu2204.json
```

adam-fowler:
What about ubuntu2404? I know swiftly doesn't yet support it but the swift.org site has downloads for them

cmcgee1024:
Yes, it appears that there are some release downloads that show up in the metadata:

```
curl -L swift.org/api/v1/install/dev/releases.json
```

```
...
    "tag": "swift-5.10.1-RELEASE",
    "xcode": "Xcode 15.3",
    "xcode_release": false,
    "date": "2024-06-05",
    "platforms": [
...
      {
        "name": "Ubuntu 22.04",
        "platform": "Linux",
        "docker": "5.10.1-jammy",
        "archs": [
          "x86_64",
          "aarch64"
        ]
      },
      {
        "name": "Ubuntu 23.10",
        "platform": "Linux",
        "docker": "5.10.1-mantic",
        "archs": [
          "x86_64",
          "aarch64"
        ]
      },
      {
        "name": "Ubuntu 24.04",
        "platform": "Linux",
        "docker": "5.10.1-noble",
        "archs": [
          "x86_64",
          "aarch64"
        ]
      },
...
```

There's no metadata for 2404 for the 6.0 (ie. main) snapshots, but the downloads page doesn't have any downloads either.

PR #153, opened 2024-08-08
Use swift.org API for getting releases and some snapshot toolchains

cmcgee1024:
See issue #149

Using the new swift.org API is more accurate than the GitHub API, also it isn't subject to the same rate limits. Use the swift.org API instead of the GitHub API for fetching metadata about the released swift toolchains, and the newer swift branches for the snapshots.

However, there is currently no metadata for snapshots on older branches than 6.0 or main. In this case use the GitHub API to get the metadata, but only for `swiftly install` and not the `swiftly list-available` commands, since the latter is very expensive when using the GitHub API.


cmcgee1024:
@swift-ci test macOS

cmcgee1024:
The macOS CI passes the tests, but there's the persistent environmental problem that reports an overall failure.

```
Test Suite 'All tests' passed at 2024-08-09 11:22:46.405.
	 Executed 67 tests, with 0 failures (0 unexpected) in 72.886 (72.895) seconds
```

Sources/Swiftly/ListAvailable.swift
```diff
@@ -28,9 +28,9 @@ struct ListAvailable: SwiftlyCommand {
         The installed snapshots for a given devlopment branch can be listed by specifying the branch as the selector:
 
             $ swiftly list-available main-snapshot
-            $ swiftly list-available 5.7-snapshot
+            $ swiftly list-available 6.0-snapshot
 
-        Note that listing available snapshots is currently unsupported. It will be introduced in a future release.
+        Note that listing available snapshots before 6.0 is unsupported.
```
adam-fowler, line 33: I assumed this is because the data isn't available at the moment. Could this change? I don't think this text is necessary. I think the response when there are no snapshots available should make it clear.

Sources/Swiftly/ListAvailable.swift
```diff
@@ -95,18 +99,11 @@ struct ListAvailable: SwiftlyCommand {
                 printToolchain(toolchain)
             }
         } else {
-            print("Available release toolchains")
+            print("Available toolchains")
             print("----------------------------")
-            for toolchain in toolchains where toolchain.isStableRelease() {
+            for toolchain in toolchains {
```
adam-fowler, line 104: Good call don't need to separate these two lists

Sources/SwiftlyCore/HTTPClient.swift
```diff
@@ -22,6 +22,100 @@ private func makeRequest(url: String) -> HTTPClientRequest {
     return request
 }
```
adam-fowler, line 24: Is it worthwhile breaking out the swift.org data and parsing into a separate file?

Sources/SwiftlyCore/HTTPClient.swift
```diff
@@ -22,6 +22,100 @@ private func makeRequest(url: String) -> HTTPClientRequest {
...
+struct SwiftOrgSwiftlyRelease: Codable {
+    var name: String
+}
+
+struct SwiftOrgPlatform: Codable {
+    var name: String
+    var platform: String
+    var archs: [String]
+
+    var platformName: String {
```
adam-fowler, line 34: Can we make it clear the difference between `platform` and `platformName`. In actual fact if we aren't using the `platform` variable lets not parse it.

Sources/SwiftlyCore/HTTPClient.swift
```diff
@@ -68,14 +162,39 @@ public struct SwiftlyHTTPClient {
...
+        let swiftOrgReleases: [SwiftOrgRelease] = try await self.getFromJSON(url: url, type: [SwiftOrgRelease].self)
+
+        var swiftOrgFiltered: [ToolchainVersion.StableRelease] = swiftOrgReleases.compactMap { swiftOrgRelease in
+            guard let swiftOrgPlatform = swiftOrgRelease.platforms.first(where: { $0.platformName == platform.name || platform.name == "xcode" }) else {
+                return nil
+            }
+
+            guard swiftOrgPlatform.archs.contains(arch) || platform.name == "xcode" else {
+                return nil
+            }
```
adam-fowler, line 193: I guess this code is just assuming there'll be a release for "xcode".

Sources/SwiftlyCore/HTTPClient.swift
```diff
@@ -88,21 +207,91 @@ public struct SwiftlyHTTPClient {
             return release
         }
 
-        return try await self.mapGitHubTags(limit: limit, filterMap: filterMap) { page in
-            try await self.getReleases(page: page)
+        guard !swiftOrgFiltered.isEmpty else {
```
adam-fowler, line 210: Is this really necessary?

adam-fowler (changes requested):
In general looks good, I've a couple of questions and minor edit requests

Sources/SwiftlyCore/HTTPClient.swift
```diff
@@ -88,21 +207,91 @@ public struct SwiftlyHTTPClient {
...
         }
     }
 
     /// Return an array of Swift snapshots that match the given filter, up to the provided
     /// limit (default unlimited).
-    ///
-    /// TODO: retrieve these directly from swift.org instead of through GitHub.
     public func getSnapshotToolchains(
+        platform: PlatformDefinition,
+        arch a: String? = nil,
```
patrickfreed, line 227: Is this arch variable just for testing? If so, then it might be better to add a method to `Platform` to retrieve the arch, which can be mocked globally rather than just in this individual function.

Sources/SwiftlyCore/HTTPClient.swift
```diff
@@ -115,8 +304,19 @@ public struct SwiftlyHTTPClient {
...
+            return []
+        }
+
+        var finalSnapshotToolchains = Array(snapshotToolchains)
+        finalSnapshotToolchains.sort(by: >)
+
+        return if let limit = limit {
+            Array(finalSnapshotToolchains[0..<limit])
+        } else {
+            finalSnapshotToolchains
```
patrickfreed, line 319: We have several variables for the snapshots, but I think we only really need `swiftOrgFiltered`. Can we drop `finalSnapshotToolchains` and `snapshotToolchains`?

Sources/Swiftly/ListAvailable.swift
```diff
@@ -45,18 +45,22 @@ struct ListAvailable: SwiftlyCommand {
...
-                SwiftlyCore.print("Listing available snapshots is not currently supported.")
-                return
+        let config = try Config.load()
+
+        let tc: [ToolchainVersion]
+
+        switch selector {
+        case let .snapshot(branch, _):
+            if case let .release(major, _) = branch, major < 6 {
+                throw Error(message: "Listing available snapshots previous to 6.0 is not supported.")
```
patrickfreed, line 55: If we're throwing an error here, we can remove the GitHub fallback.

cmcgee1024: The GitHub fallback is still there in case someone wants to directly `swiftly install` a snapshot from pre-6.0 branch. They just can't list them using this subcommand.

Sources/Swiftly/ListAvailable.swift
cmcgee1024: This else only gets invoked in the absence of a selector, which defaults to all of the releases, since every snapshot build for every branch would be overwhelming and probably not very useful.

I can retain the "release" toolchains label to make it clear that these are only releases, but the where clause shouldn't be necessary here.

Sources/SwiftlyCore/HTTPClient.swift
cmcgee1024: I'm on the fence about this.

It feels like the swift.org metadata is a foundational use of the swiftly http client, so why not just put the code there? On the other hand, because of the GitHub fallback there's some asymmetry now because it's in another source file. But, there's also a feeling that the fallback may go away someday, so that could be why it is separate.

Sources/SwiftlyCore/HTTPClient.swift
cmcgee1024: The platform is not necessary here. It'll be removed in the latest revision. Also, the platformName gets a comment indicating it's purpose, which is to match swiftly's PlatformDefinition.name field.

Sources/SwiftlyCore/HTTPClient.swift
cmcgee1024: In each SwiftOrgRelease there is a list of platforms with their archs. The exception is xcode, which doesn't have an entry in that platform list. I'll see if I can rewrite this to be clearer.

Sources/SwiftlyCore/HTTPClient.swift
cmcgee1024: Agreed, this is not needed. I'll remove it.

Sources/SwiftlyCore/HTTPClient.swift
cmcgee1024: Right, this allows the HTTPClientTests to be able to test for all os/arch, independently from the platform that is currently running the test. I don't think that we need a mock for it since most of the tests rely on the current platform, and could run into problems if they try to install a toolchain for an alternate platform.

Sources/SwiftlyCore/HTTPClient.swift
cmcgee1024: I'm looking into making this simpler and easier to follow.

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
The macOS CI passed the xctests despite reporting an overall failure:

```
Test Suite 'All tests' passed at 2024-08-20 16:11:50.665.
	 Executed 67 tests, with 0 failures (0 unexpected) in 73.780 (73.789) seconds
```

Sources/SwiftlyCore/HTTPClient.swift
adam-fowler: As I see it there is the client and the usage of the client which are two different things

cmcgee1024:
@patrickfreed @adam-fowler how do these latest changes look? Any remaining issues?

adam-fowler (approved):

Sources/SwiftlyCore/HTTPClient.swift
```diff
@@ -68,14 +163,42 @@ public struct SwiftlyHTTPClient {
...
-        let filterMap = { (gh: GitHubTag) -> ToolchainVersion.StableRelease? in
-            guard let release = try gh.parseStableRelease() else {
+        let arch = if a == nil {
+#if arch(x86_64)
+            "x86_64"
+#elseif arch(arm64)
+            "aarch64"
+#else
+            #error("Unsupported processor architecture")
+#endif
```
patrickfreed, line 179: we do this in a few places, can we extract this to a method/computed variable on `Platform`? I think it would also be good to define an enum for it so we don't have to rely on magic strings here and elsewhere.

Tests/SwiftlyTests/HTTPClientTests.swift
```diff
@@ -45,35 +45,38 @@ final class HTTPClientTests: SwiftlyTests {
...
-        // WHEN: we ask for the first page of releases with page size 5
-        var releases = try await SwiftlyCore.httpClient.getReleases(page: 1, perPage: 5)
-        // THEN: we get five releases
-        XCTAssertEqual(5, releases.count)
+    func testGetMetdataFromSwiftOrg() async throws {
+        let supportedPlatforms = [
+            PlatformDefinition.macOS,
+            PlatformDefinition.ubuntu2204,
+            PlatformDefinition.ubuntu2004,
+            // PlatformDefinition.ubuntu1804,
```
patrickfreed, line 53: why commented out?

Sources/SwiftlyCore/HTTPClient.swift
```diff
@@ -88,35 +211,103 @@ public struct SwiftlyHTTPClient {
...
+        } else {
+            a!
+        }
+
+        let branchLabel = switch branch {
+        case .main:
+            "main"
+        case let .release(major, minor):
+            "\(major).\(minor)"
+        }
```
patrickfreed, line 270: `branch.name` gives us this.

Sources/SwiftlyCore/HTTPClient.swift
```diff
@@ -68,14 +163,42 @@ public struct SwiftlyHTTPClient {
...
+#endif
+        } else {
+            a!
+        }
+
+        let url = "https://swift.org/api/v1/install/releases.json"
+        let swiftOrgReleases: [SwiftOrgRelease] = try await self.getFromJSON(url: url, type: [SwiftOrgRelease].self)
+
+        var swiftOrgFiltered: [ToolchainVersion.StableRelease] = swiftOrgReleases.compactMap { swiftOrgRelease in
+            if platform.name != "xcode" {
```
patrickfreed, line 188: `"xcode"` -> `PlatformDefinition.xcode.name`

Sources/SwiftlyCore/HTTPClient.swift
```diff
@@ -68,14 +163,42 @@ public struct SwiftlyHTTPClient {
...
+                guard let swiftOrgPlatform = swiftOrgRelease.platforms.first(where: { $0.platformName == platform.name }) else {
+                    return nil
+                }
+
+                guard swiftOrgPlatform.archs.contains(arch) else {
+                    return nil
+                }
+            }
+
+            guard let version = try? ToolchainVersion(parsing: swiftOrgRelease.stableName),
```
patrickfreed, line 199: if parsing fails, I think we'd want to let the user know about it in some way, either just by throwing here (which I think is fine, since we should never fail parsing swift.org results) or logging.

Sources/SwiftlyCore/HTTPClient.swift
```diff
@@ -88,35 +211,103 @@ public struct SwiftlyHTTPClient {
...
+        }
+
+        let branchLabel = switch branch {
+        case .main:
+            "main"
+        case let .release(major, minor):
+            "\(major).\(minor)"
+        }
+
+        let platformName = if platform.name == "xcode" {
```
patrickfreed, line 272: `"xcode"` -> `PlatformDefinition.xcode.name`, also below.

Sources/SwiftlyCore/HTTPClient.swift
```diff
@@ -22,6 +22,101 @@ private func makeRequest(url: String) -> HTTPClientRequest {
...
+    var name: String
+}
+
+struct SwiftOrgPlatform: Codable {
+    var name: String
+    var archs: [String]
+
+    /// platformName is a mapping from the 'name' field of the swift.org platform object
+    /// to swiftly's PlatformDefinition name.
+    var platformName: String {
```
patrickfreed, line 35: To better align with this var's usage/documentation, it might be better to just return the matching `PlatformDefinition` or `nil` if it isn't supported. That way we can rely on the `PlatformDefinition` constants as the single source of supported platform names, and we also don't need to define constants for platforms we don't support. If we go that direction, renaming this something like `asPlatformDefinition` or something might make sense.

If we don't change this in that way, then I think at least renaming it to `canonicalPlatformName` or `swiftlyPlatformName` communicates the difference between the other `name` field a bit better. We'd also want to use the `PlatformDefinition` name constants here as well.

Sources/SwiftlyCore/HTTPClient.swift
```diff
@@ -88,35 +211,103 @@ public struct SwiftlyHTTPClient {
...
+
+        let branchLabel = switch branch {
+        case .main:
+            "main"
+        case let .release(major, minor):
+            "\(major).\(minor)"
+        }
+
+        let platformName = if platform.name == "xcode" {
+            "macos"
```
patrickfreed, line 273: the inconsistency between these swift.org APIs and swift.org's download URLs here is a bit odd. Is there any chance the API would be updated to use "xcode" like the downloads do so we can avoid needing to adjust for it here?

This question also applies for the non-macOS platform names. It'd be nice if e.g. `ubuntu1804` were used rather than the pretty name.

patrickfreed (approved):
Overall looks good, super excited to see swiftly using official APIs now. I have a bunch of smaller, mostly style suggestions. Feel free to merge after addressing without waiting for another round of feedback from me.

Sources/Swiftly/ListAvailable.swift
```diff
@@ -45,18 +45,22 @@ struct ListAvailable: SwiftlyCommand {
...
             }
-        }
 
-        let toolchains = try await SwiftlyCore.httpClient.getReleaseToolchains()
-            .map(ToolchainVersion.stable)
-            .filter { selector?.matches(toolchain: $0) ?? true }
+            tc = try await SwiftlyCore.httpClient.getSnapshotToolchains(platform: config.platform, branch: branch).map { ToolchainVersion.snapshot($0) }
+        default:
```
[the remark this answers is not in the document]
cmcgee1024: Thanks, I've changed this to add explicit stable/latest case, and a default to catch no selector provided.

Sources/SwiftlyCore/HTTPClient.swift
cmcgee1024: This pattern is isolated to two places at the moment, mostly isolated to path/filename calculation and internal swift org metadata. It might be possible to isolate all of it in the HTTPClient, or an HTTPClient+SwiftOrg.

There are some problems with pulling this out into a global enum at the Platform level. In macOS, the arch isn't used at all because of the universal binaries/packages. Also, there is some disagreement between the naming of the 64-bit arm architecture. In some places. it's "arm64" while others is "aarch64." It doesn't feel like an enum could be universally defined across OSes and cases without being a leaky abstraction.

I feel that in Swift it's quite common to put these kinds of macros (os/arch) at the place where they are needed as long as it isn't excessive and it can maintain locality of change when/if the logic changes. For now, it's only these two places: HTTPClient, and Install.

Tests/SwiftlyTests/HTTPClientTests.swift
cmcgee1024: Ubuntu 18.04 doesn't have toolchains available for the releases that we are testing here. I'll update the comment to explain.

Sources/SwiftlyCore/HTTPClient.swift
cmcgee1024: Thanks, this is a really nice pattern here. We can return the constants defined in PlatformDefinition for platforms that swiftly supports, and manually constructed PlatformDefinitions for the ones that it doesn't, which can provide enough metadata to represent what the swift.org API is producing. For yet unknown (to swiftly) swift.org platforms, then it's just nil until this code adds a representation for it.

Sources/SwiftlyCore/HTTPClient.swift
cmcgee1024: @shahmishal here is some feedback regarding the swift.org API. There are some inconsistencies between the new API and the toolchain download URL's / filenames. We're working around these for now in swiftly, but perhaps this can be made a little more ergonomic for other consumers of the API.

cmcgee1024:
@swift-ci test macOS

Issue #22, opened 2022-11-11
Implement `exec` command


patrickfreed:
swiftly should have an `exec` command which allows a user to run a specific `swift` command without having to switch the active toolchain.

e.g.
```
swiftly exec 5.5 build
```
Would run `swift build` using the latest 5.5 toolchain.

cmcgee1024:
Here are some additional thoughts on this enhancement. First, the version selector can be optional, and then the toolchain that is in use would be chosen, where the used toolchain logic can be flexible, possibly consulting the `.swift-version` file (see #155) as well as the global default in the `config.json`.

At the expense of requiring the full toolchain command there are some additional features we can unlock. Here's what the build example above could look like:
```
swiftly exec 5.5 swift build
```

But now, you aren't restricted to just the `swift` command from the toolchain. You can also invoke any of the other tools, such as lldb, and clang too using the in-use toolchain:
```
swiftly exec clang -o foo.o foo.c
```

Or, you can invoke a command or even a shell script that makes use of any tool in the toolchain because it sets the toolchain at the beginning of the PATH:
```
swiftly exec ./my-special-script.sh
```

And, if your project uses autoconf/cmake you can build it with the swift toolchain (clang, clang++) like this because exec sets the CC, CXX environment variables to the in-use toolchain:
```
swiftly exec ./configure.sh
swiftly exec make
```

cmcgee1024:
This got implemented along with the proxy support in the form of a `swiftly run` subcommand. Note that one needs to provide the command to run since it supports running arbitrary commands, such as make/clang/swift.

PR #155, opened 2024-08-10
Swiftly proxies

cmcgee1024:
Add a design for the new swiftly proxy system for more flexible routing of command invocations to a toolchain.

Re-target the symbolic links that swiftly creates to point to swiftly itself, which can decide the selected toolchain more flexibly on a per-invocation basis.

Add support for reading the toolchain version from a `.swift-version` file if a selector isn't given. The `swiftly use` command will attempt to update a swift version file, or in certain circumstances it will create the version file for you. The special `swiftly install` with no selector argument that will attempt to install a toolchain version specified in a version file.

Create a `swiftly run` subcommand that can run arbitrary commands, such as build scripts or makefile generators. This subcommand will set the PATH environment variable to the currently selected toolchain, and other special environment variables like CC and CXX to specifically point to the clang and clang++ executables. Add an ability for this subcommand to accept arbitrary toolchain selectors that take the higher precedence over the presence of the swift version file or the global default.

```
swiftly run swift build +latest
swiftly run cmake -G ninja +main-snapshot
```

Update `swiftly use` and `swiftly list` to take into account that there is a global default, which is the fallback if there is no `.swift-version` file, but also a potentially different toolchain that is in use in the current context. The `swiftly use` subcommand with not arguments will show in brackets the reason why the current in-use toolchain is selected, whether it is a special `.swift-version` file in which case it will show the file path, or it is the global default.


BrianHenryIE:
Hijacking`swift` altogether isn't ideal. I've used `which swift` to determine the toolchain directory and proxying adds a level of indirection that isn't obvious.

https://github.com/anders0nmat/swift-embedded-pm/pull/3/files

Consider:

```
swiftly run 5.10.1 build
```

Or proxy CLI arguments that don't exist in swiftly to `swift`:

```
swiftly build
```

(I just write Swift for fun, so take my suggestions with a grain of salt)

cmcgee1024:
@BrianHenryIE your feedback is appreciated, thank you.

> I've used `which swift` to determine the toolchain directory and proxying adds a level of indirection that isn't obvious.
> 
> https://github.com/anders0nmat/swift-embedded-pm/pull/3/files

It seems from the above patch that the end goal is to be able to invoke the llvm-ar executable from the current toolchain. I'm curious, why dig through the symlink to the real path using the `realpath` command? The `llvm-ar` command should be a sibling of the `swift` command, whether if it is a symlink created by swiftly, or a toolchain binary installed by other means (e.g. manual installation, system package). Inspecting the symlink would seem to couple this CMake script to swiftly being the toolchain manager.

> Consider:
> 
> ```
> swiftly run 5.10.1 build
> ```

Providing a swiftly run subcommand is something that we can consider as an enhancement to the current swiftly design. We could allow someone to run any command that they want, not just the toolchain ones. The command runs where the PATH is set to the toolchain binaries with the highest precedence. Also, we can set certain environment variables like CC and CXX to the selected clang version.

```
# Let's use the swift toolchain to compile this C project:
swiftly run 5.10 ./configure
swiftly run 5.10 make
```

We can put these ideas and more into #22

DESIGN.md
```diff
@@ -208,6 +216,10 @@ To use the latest installed main snapshot, leave off the date:
 
 `swiftly use main-snapshot`
 
+The use subcommand also supports `.swift-version` files. If version file is present in the current working directory, or an ancestory directory, then swiftly will update that file with the new version to use. This can be a useful feature for a team to share and align on toolchain versions with git. As a special case, if swiftly could not find a version file, but it could find a Package.swift file it will create a new version file for you in the package and set that to the requested toolchain version.
+
+Note: The `.swift-version` file mechanisms can be overridden using the `--global-default` flag so that your swiftly installation's default toolchain can be set explicitly.
```
adam-fowler, line 221: I would rather the default is the other way. ie `use` edits global settings and you can change local settings with a `--local` flag.

DESIGN.md
```diff
@@ -266,6 +278,18 @@ This command checks to see if there are new versions of `swiftly` itself and upg
...
+
+Swiftly will create a set of symbolic links in its SWIFTLY_BIN_DIR during installation that point to the swiftly binary itself for each of the common toolchain commands, such as swift, swiftc, clang, etc. This mechanism will allows swiftly to proxy those command invocations to a requested toolchain. A toolchain can be requested in these ways in order of precedence:
+
+* Special toolchain selectors among the regular tool command-line arguments (e.g. `swift build +5.10.1`) with the special '+' prefix
+* The presence of a .swift-version file in the current working directory, or ancestor directory, with the required toolchain version
+* The swiftly default (in-use) toolchain set in the config.json by `swiftly install` or `swiftly use` commands
+
+In the first two cases, if there is no matching toolchain installed, swiftly will attempt to automatically install the requested toolchain and use it if the installation succeeeds.
+
+Note: If swiftly automatically installs a toolchain during proxying and that toolchain requires post installation steps then the proxy will abort with those post installation instructions.
```
adam-fowler, line 291: Note: IDEs like VSCode need to know the actual path to the swift executable (currently VSCode follows the symbolic links) so it can use other components of the selected toolchain. We might need to add a new subcommand to provide this information.

adam-fowler (commented):
I'm not 100% convinced by this yet. I wonder how the core team would feel about the swift exe being hijacked by swiftly.

DESIGN.md
cmcgee1024: Ah, so this is why my vscode experiments with swiftly have failed so far. Can you provide more details of what sorts of details (components) the vscode needs from the toolchain? Maybe we can abstract that like we do the location of the toolchain executables in swiftly.

cmcgee1024:
> I'm not 100% convinced by this yet. I wonder how the core team would feel about the swift exe being hijacked by swiftly.

I'm definitely interested to hear if there are opinions from the core team about this enhancement. I prefer to think of this as extending the semantics, not "hijacking" the toolchain itself. There are precedents for a proxy mechanism of a toolchain manager on top of the toolchain itself, such as `rustup`.

DESIGN.md
cmcgee1024: I think that there are advantage of having this local as the default. It encourages the sharing of the toolchain revision that the team is working with through discovery of this feature of swiftly. Also, this helps to tie into one of the values of using swiftly itself, which is how easily and flexibly you can switch toolchains.

In terms of precedence, git defaults to local over global for its configuration.

I'm curious what you think are the pitfalls of this approach.

DESIGN.md
adam-fowler: All we need is the path to the toolchain. It then uses this to find various executables (lldb-dap, sourcekit-lsp, llvm-cov) which I can see being abstracted, we also need the path to lldb dynamic library to pass to CodeLLDB. Not sure how you'd do that. There might be other things I cannot remember offhand.

If we add a subcommand to provide this information, we would also need to edit the extension of course to use this information.

cmcgee1024: Thanks, I can try and dig deeper about the lldb dylib issue.

It seems to me that swiftly is basically an abstract toolchain, and should be able to cleverly pretend that it is a static toolchain in most (maybe all) usages. We could provide a subcommand to give use the real toolchain location, but it feels like an escape hatch from the abstraction. I'd like to understand why swiftly couldn't provide the complete abstraction.

DESIGN.md
```diff
@@ -266,6 +278,18 @@ This command checks to see if there are new versions of `swiftly` itself and upg
...
 
+### Proxy command invocations to a requested toolchain
+
+Swiftly will create a set of symbolic links in its SWIFTLY_BIN_DIR during installation that point to the swiftly binary itself for each of the common toolchain commands, such as swift, swiftc, clang, etc. This mechanism will allows swiftly to proxy those command invocations to a requested toolchain. A toolchain can be requested in these ways in order of precedence:
+
+* Special toolchain selectors among the regular tool command-line arguments (e.g. `swift build +5.10.1`) with the special '+' prefix
+* The presence of a .swift-version file in the current working directory, or ancestor directory, with the required toolchain version
+* The swiftly default (in-use) toolchain set in the config.json by `swiftly install` or `swiftly use` commands
+
+In the first two cases, if there is no matching toolchain installed, swiftly will attempt to automatically install the requested toolchain and use it if the installation succeeeds.
```
adam-fowler, line 289: We'll need a way to disable this. People with a bad network connection will not be wanting to download 900MB of swift toolchain just to build a project.

cmcgee1024: Yes, I'll update this since this has come up a few times. It could be a case of over automation. Instead, if there's no matching toolchain then the command fails. The `swiftly install` with no arguments can be used to install the currently selected toolchain with the full capabilities of that subcommand, including post install script, etc. Also, it's clear that an error is related to the install operation itself since that is the request.

DESIGN.md
cmcgee1024: Ok, I see now that there are cases where swiftly's toolchain abstraction isn't going to be sufficient. I've elaborated on the boundaries of swiftly's abstraction: just the toolchain binaries, and not for long-lived invocations. There's an escape hatch through a special `swiftly use --location` flag that will provide the physical location of a toolchain, and also help for changes in toolchain selection. It does mean that if tools operate outside the boundaries then they might need to be aware of swiftly and interface with it to provide the best user experience.

adam-fowler: I'm not looking to override the location of the toolchain. I'm looking for the location of the  current toolchain.

cmcgee1024: If you invoke `swiftly use` without a selector today it will tell you the version number of the toolchain that is in use in the current context. I'm proposing the addition of the `--location` flag here to make that output the file path of the toolchain that is in use. It probably wouldn't make sense for the flag to be used when you provide a selector, so that can be checked in the subcommand and documented.

DESIGN.md
adam-fowler: Could we also have a command line option to ignore the `.swift-version` file. This would make it easier for things like CI to test a library (which includes a `.swift-version` file) against multiple swift toolchains

cmcgee1024: Yes, I think that this in this case the command-line selectors will take precedence over both the .swift-version file and the global default in-use toolchain from the config.json. The CI can do this to force specific toolchain(s):

```
swiftly install 5.10.1 # Install this specific toolchain
swift build +5.10.1   # Build with this toolchain
```

0xTim (commented):
I'd be interested to hear the core team's view on intercepting any calls to `swift`. Node does a similar thing with `nvm` IIRC so there's precedent, not that that's the best example of an ecosystem if there are any security concerns 😅 

I would say that the `+` syntax is not overly discoverable or intuitive, I'm not aware of any other tool that has this. I'd be more inclined to have an actual flag that can either be intercepted or called via `swiftly run` etc.

Overall however I agree that having a mechanism for being able to run commands in Swift with different versions without having to constantly switch your local version is valuable and should be implemented

cmcgee1024:
> I'd be interested to hear the core team's view on intercepting any calls to `swift`. Node does a similar thing with `nvm` IIRC so there's precedent, not that that's the best example of an ecosystem if there are any security concerns 😅

Do you know a way to engage the core team on this? I'm also curious if there are any concerns or suggestions from there. The proxy idea has been mentioned in the forums a couple of times now, the last one points to this PR. I have received some feedback, but I don't know if any have been representative of the core.

> I would say that the `+` syntax is not overly discoverable or intuitive, I'm not aware of any other tool that has this. I'd be more inclined to have an actual flag that can either be intercepted or called via `swiftly run` etc.

Thank you, this is a fair observation.

The proxy concept along with the `+` syntax is inspired in part by `rustup`:
=> https://rust-lang.github.io/rustup/concepts/proxies.html

In their documentation there is only a small mention of the special selector syntax:
=> https://rust-lang.github.io/rustup/overrides.html#toolchain-override-shorthand

I think that this syntax can be discoverable in swiftly by providing (some) examples in the getting started guide, demonstrating how to use swiftly. In the swiftly steady state I imagine that the syntax won't be used very often as the toolchain will likely be selected with either the in-use global default for the user, or hopefully, the project's `.swift-version` file that aligns the project participants. In these cases swiftly works just like using the toolchain directly, with the exception of times when the selected toolchain isn't installed where it prompts the user to install it. This is what's interesting about the proxy concept.

I can see the `+` syntax being particularly useful when someone wants to run a check against the latest snapshot, or a previous stable build.

> Overall however I agree that having a mechanism for being able to run commands in Swift with different versions without having to constantly switch your local version is valuable and should be implemented

I'm wondering, since this has come up a few times, if it might be worth providing both proxy capability and #22 at the same time in this PR to suit different preferences and use cases.

DESIGN.md
```diff
@@ -266,6 +285,28 @@ This command checks to see if there are new versions of `swiftly` itself and upg
...
+If swiftly cannot find an installed toolchain that matches the selection then it fails with an error and instructions how to use `swiftly install` to satisfy the selection next time.
+
+A few notes about the '+' prefix. First, if a literal '+' prefix should be sent directly to the tool as an argument then it is escaped by doubling it with '++'. An argument with only '++++' is ignored entirely, but any additional arguments are sent directly to the tool without any further inspection of their prefixes. This is analogous to the special '--' token that certain argument parsers accept so that they don't interpret anything following that token as command flags or options.
+
+#### Resolve selected toolchain
+
+For cases where the physical toolchain must be located, such as references specific header files, or shared libraries that are not proxied by swiftly there is a method to resolve the currently selected toolchain to its physical location using `swiftly use`.
+
+```
+swiftly use --location
```
patrickfreed, line 305: perhaps `--print-location`? As it is, it looks like `--location` affects `use` in some way.

DESIGN.md
```diff
@@ -60,7 +67,7 @@ A simple setup for managing the toolchains could look like this:
 
 The toolchains (i.e. the contents of a given Swift download tarball) would be contained in the toolchains directory, each named according to the major/minor/patch version. `config.json` would contain any required metadata (e.g. the latest Swift version, which toolchain is selected, etc.). If pulling in Foundation to use `JSONEncoder`/`JSONDecoder` (or some other JSON tool) would be a problem, we could also use something simpler.
 
-The `~/.local/bin` directory would include symlinks pointing to the `bin` directory of the "active" toolchain, if any.
+The `~/.local/bin` directory would include symlinks pointing to swiftly itself. When these binaries are run swiftly proxies them to the requested toolchain, or a default.
```
patrickfreed, line 70: I think it would be good to more clearly explain what "these binaries" refer to in this sentence.

patrickfreed (commented):
I personally tend to prefer a subcommand over the `+` syntax since swiftly isn't ubiquitous (yet :slightly_smiling_face:) like rustup is, and so introducing swiftly-specific syntax to a non-swiftly binary invocation may cause confusion for non-swiftly users when they encounter someone else using. That said, it is a generally useful syntax, so if the core team is supportive, then I don't have a problem with it.

To avoid blocking the work for this, I'd suggest we start with the more conservative `swiftly run <toolchain>` or `swiftly exec <toolchain>` command and then add the `+` syntax later.

0xTim:
> Do you know a way to engage the core team on this?

Maybe @shahmishal can bring it up but this should probably be raised at the Platform Steering Group until whatever group will end up with ownership of this takes over them

airspeedswift:
> this should probably be raised at the Platform Steering Group

I think this is right. There are plans to add a third steering group in future (dubbed "Ecosystem") and this might fall under that rather than Platform, but I could see it going either way in this case so starting with the platform steering group is the right way to go for now.

The first step is to write up a pitch and post it to the forums. That pitch should cover the pros/cons of the approach, and recommend one (with the others being alternatives considered). Then engagement on the merits can happen in that thread. That said, it's fine to hash out early ideas on a doc in this repo beforehand.

Once the pitch has had some time to converge on a way forward, the platform steering group would take it up as a proposal. You don't need the steering group to pre-clear that pitch, though you might want to give them a heads up about it, so they can engage in it too since they'll ultimately be the ones deciding which way to go with this.

Note that since Swiftly was ingested into swiftlang and so is part of the Swift project now, I don't see anything contentious about close integration between the swift launch tool and swiftly per se. But of course, the other tradeoffs cited in this thread about exactly how that integration actually works should be discussed.

cmcgee1024:
Thanks everyone for the discussion here. This has been really positive. I will change the shape of this PR in a couple of ways and see if this can be accomplished in the near term:

* Retain proxy concept for further refinement in the future.
* Implement #22 in here to support a swiftly specific syntax for on-the-fly toolchain selection and also the extra capabilities that feature can provide.
* Defer the special toolchain selector syntax (ie. the injection of the '+' into tool invocations) for later after a pitch has been made on the forums and the platform steering group.

Please let me know if there are any concerns over this latest pivot.

cmcgee1024:
@swift-ci test macOS

rauhul:
Out of curiosity, do you have an idea of how this would interact with xcrun?

cmcgee1024:
> Out of curiosity, do you have an idea of how this would interact with xcrun?

Swiftly's mechanisms for toolchain selection are entirely separate from xcrun. But, it will install toolchain packages in the user's volume, and those should be reachable with xcrun.

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
The macOS check is passing despite the failure:

```
Test Suite 'All tests' passed at 2024-09-09 17:12:09.524.
	 Executed 68 tests, with 0 failures (0 unexpected) in 61.581 (61.592) seconds
```

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
macOS is still passing all of the tests:

```
Test Suite 'All tests' passed at 2024-09-10 11:04:46.361.
	 Executed 68 tests, with 0 failures (0 unexpected) in 61.283 (61.293) seconds
```

Sources/Swiftly/Proxy.swift
```diff
@@ -0,0 +1,59 @@
...
+    "clangd",
+    "swift",
+    "docc",
+    "swiftc",
+    "lld",
+    "llvm-ar",
+    "plutil",
+    "repl_swift",
+    "wasm-ld",
+]
```
patrickfreed, line 21: I'm a bit concerned about this--what if a new toolchain has a new executable not on this list? Or what if a toolchain doesn't have a executable on this list? Perhaps being concerned about that is too pessimistic, but it seems somewhat brittle to me. I can't think of any solutions off the top of my head given that we want to support selecting a toolchain via `.swift-version` files, though. The only thing I can think of is requiring users to run `swiftly use` to select the toolchain from the file, but that does seem quite un-ergonomic.

patrickfreed (commented):
Sorry for the delay!

Documentation/SwiftlyDocs.docc/swiftly-cli-reference.md
```diff
@@ -424,3 +438,61 @@ swiftly self-update [--version] [--help]
...
+The script will receive the argument as '+abcde'. If there are multiple arguments with the '+' prefix that should be escaped you can disable the selection using a '++' argument, which turns off any selector argument processing for subsequent arguments. This is anologous to the '--' that turns off flag and option processing for subsequent arguments in many argument parsers.
+
+    $ swiftly run ./myscript.sh ++ +abcde +xyz
+
+The script will receive the argument '+abcde' followed by '+xyz'.
+
+
+**--version:**
+
+*Show the version.*
```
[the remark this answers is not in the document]
cmcgee1024: This is a global flag coming from the argument parser, not an explicit option for the run subcommand. See the `--help` below. The CLI reference generator that I wrote for swiftly could probably use a bit of refinement to skip over these special flags. It's very naive at the moment, but I think better than having to maintain our own reference manually.

DESIGN.md
```diff
@@ -266,6 +285,58 @@ This command checks to see if there are new versions of `swiftly` itself and upg
...
+swiftly run swift build +5.10.1 # Runs swift build with the 5.10.1 toolchain
+```
+
+A few notes about the '+' prefix. First, if a literal '+' prefix should be sent directly to the tool as an argument then it is escaped by doubling it with '++'. An argument with only '++' is ignored entirely, and any additional arguments are sent directly to the command without any further inspection of their prefixes. This is analogous to the special '--' token that certain argument parsers accept so that they don't interpret anything following that token as command flags or options.
+
+If the selected toolchain is not installed then swiftly will exit with a message indicating that you need to run `swiftly install x.y.z` to install it. However, if you enter a special `+install` token then swiftly will automatically download and install the toolchain if it isn't already present.
+
+```
+# Download and install the latest main snapshot toolchain and run 'swift build' to build the package with it.
+swiftly run swift build +main-snapshot +install
```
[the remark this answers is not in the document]
cmcgee1024: Thank you, I'll remove this. It might have made more sense in the context of the original proxy proposal.

Sources/Swiftly/Use.swift
```diff
@@ -46,43 +53,187 @@ internal struct Use: SwiftlyCommand {
...
+                break
+            }
+
+            let svFile = cwd.appendingPathComponent(".swift-version")
+
+            if FileManager.default.fileExists(atPath: svFile.path) {
+                let contents = try? String(contentsOf: svFile, encoding: .utf8)
+
+                guard let contents = contents else {
+                    return (nil, .swiftVersionFile(svFile, nil, Error(message: "The swift version file could not be read: \(svFile)")))
```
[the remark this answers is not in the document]
cmcgee1024: The error is part of the tuple here because in some contexts the error is ignored, while others it is thrown.

For example, when `swiftly use` is displaying the currently selected toolchain, or it is trying to find the swift version file to update with the used version it doesn't care if the file is well-formed, or not matching a toolchain that is currently installed. Whereas in other cases like the proxies or the `swiftly run` that will interfere with the ability to run the command against a toolchain.

Sources/Swiftly/Use.swift
```diff
@@ -46,43 +53,187 @@ internal struct Use: SwiftlyCommand {
...
+            }
         }
 
-        guard try Swiftly.currentPlatform.use(toolchain, currentToolchain: previousToolchain) else {
-            return
+        if case let .swiftVersionFile(versionFile, _, _) = result {
+            // We don't care in this case if there were any problems with the swift version files, just overwrite it with the new value
+            try toolchain.name.write(to: versionFile, atomically: true, encoding: .utf8)
+        } else if let newVersionFile = findNewVersionFile(), !globalDefault {
+            try toolchain.name.write(to: newVersionFile, atomically: true, encoding: .utf8)
```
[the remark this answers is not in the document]
cmcgee1024: Thanks, I can change this over to a prompt to either go ahead and create a file that can be shared, or update the global default.

Sources/Swiftly/Proxy.swift
cmcgee1024: The idea of the stable proxy allowed list came out of some necessity, which was to install the proxies at swiftly init time instead of toolchain installation time, but also as a measure of security, which strongly favours allow lists over deny lists.

I can imagine that the commands in this list can be useful to someone in the context of swiftly. We can change this list over time, and manage the proxy symlinks as we do upgrades from version to version. If anyone has suggestions on adding/removing commands from the list I'm interested in hearing your opinions.

This is also one of the reasons that I updated the swiftly design document to clarify that the purpose of swiftly is not to be a fully virtual toolchain. Not all clients of the toolchain will be able to work within the abstraction that swiftly provides. In one case there is a need to reach inside and grab a dylib (.so). There are potentially many others where someone would like to reach in for a particular header file, or other resources. Swiftly isn't equipped to symlink, or proxy, everything in there. It's just not practical.

So, for cases where someone needs access to something more obscure, there is the escape hatch of the `swiftly use --print-location` to get at the real toolchain path where they can reach into the internals as needed at the cost of being optionally dependent on swiftly for that capability.

Documentation/SwiftlyDocs.docc/swiftly-cli-reference.md
```diff
@@ -424,3 +438,61 @@ swiftly self-update [--version] [--help]
 
 
 
+## run
+
+Run a command while proxying to the selected toolchain commands.
+
+```
+swiftly run <command>... [--version] [--help]
```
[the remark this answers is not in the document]
cmcgee1024: This documentation is generated using a plugin that inspects the argument parser. The `+` syntax is specific to this subcommand and so the usage here doesn't pick this up as a typical flag, option, or argument. Hopefully the examples usages below capture how to use this syntax. Note that the `+` isn't strictly required, so not knowing about it doesn't block naive uses of `swiftly run`.

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
The RHEL 9 tests failed on an network failure, otherwise everything passed.

DESIGN.md
```diff
@@ -138,6 +145,14 @@ Installing a specific snapshot from a swift version development branch
 
 `swiftly install 5.5-snapshot-2022-1-28`
 
+##### Installing the version from the `.swift-version` file
+
+A package could have a swift version file that specifies the recommended toolchain version. A swiftly install with no version will search for a version file and install that version.
+
+`swiftly install`
+
+If no swift version file can be found then the installation fails indicating that it couldn't fine the file.
```
justice-adams-apple, line 154: Very small nit, but could we use
> If no .swift-version file ...

Rather than 
> swift version file

Just to be consistent, so users don't conflate the "swift version file" with something else in the Swift ecosystem. This way we explicitly mention the file in question.  This is how it's written in the _"Installing the version from the `.swift-version` files"_ section

DESIGN.md
```diff
@@ -208,6 +223,10 @@ To use the latest installed main snapshot, leave off the date:
 
 `swiftly use main-snapshot`
 
+The use subcommand also supports `.swift-version` files. If version file is present in the current working directory, or an ancestory directory, then swiftly will update that file with the new version to use. This can be a useful feature for a team to share and align on toolchain versions with git. As a special case, if swiftly could not find a version file, but it could find a Package.swift file it will create a new version file for you in the package and set that to the requested toolchain version.
```
justice-adams-apple, line 226:
```suggestion
If a .swift-version file 
```
Simple wording change suggestion

Documentation/SwiftlyDocs.docc/swiftly-cli-reference.md
```diff
@@ -53,6 +53,10 @@ Likewise, the latest snapshot associated with a given development branch can be
     $ swiftly install 5.7-snapshot
     $ swiftly install main-snapshot
 
+ Install whatever toolchain is current selected, such as a .swift-version file:
```
justice-adams-apple, line 56:
```suggestion
 Install whatever toolchain is currently selected, such as a .swift-version file:
```

Documentation/SwiftlyDocs.docc/swiftly-cli-reference.md
```diff
@@ -424,3 +443,59 @@ swiftly self-update [--version] [--help]
...
+```
+swiftly run <command>... [--version] [--help]
+```
+
+**command:**
+
+*Run a command while proxying to the selected toolchain commands.*
+
+
+Run a command with a selected toolchain, so that all toolchain commands are become the default added to the system path and other common environment variables.
```
justice-adams-apple, line 459: The wording of this sentence confuses me a bit, might just need some reworking

Sources/Swiftly/Install.swift
```diff
@@ -39,9 +39,13 @@ struct Install: SwiftlyCommand {
 
             $ swiftly install 5.7-snapshot
             $ swiftly install main-snapshot
+
+         Install whatever toolchain is current selected, such as a .swift-version file:
```
justice-adams-apple, line 43:
```suggestion
         Install whatever toolchain is currently selected, such as a .swift-version file:
```

Sources/Swiftly/Uninstall.swift
```diff
@@ -100,10 +100,9 @@ struct Uninstall: SwiftlyCommand {
                     ?? config.listInstalledToolchains(selector: .latest).filter({ !toolchains.contains($0) }).max()
                     ?? config.installedToolchains.filter({ !toolchains.contains($0) }).max()
                 {
-                    try await Use.execute(toUse, &config)
+                    try await Use.execute(toUse, globalDefault: true, &config)
                 } else {
                     // If there are no more toolchains installed, just unuse the currently active toolchain.
-                    try Swiftly.currentPlatform.unUse(currentToolchain: toolchain)
```
justice-adams-apple, line 106: Do we still want to `unuse` the toolchain if there are no more toolchains to be set active, in order to clean up all the symlinks?

cmcgee1024: With this change the proxy symlinks are created at install time of swiftly and remain there. Keeping them stable helps with certain shells that aggressively cache commands on the PATH, such as zsh. If there are no toolchains installed and someone tries to run "swift", "clang", or some other toolchain command then swiftly will give an error indicating that they can install one.

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
Note that macOS CI is failing in a way that is unrelated to the tests. The tests are all green:

```
Test Suite 'All tests' passed at 2024-10-18 12:48:47.119.
	 Executed 67 tests, with 0 failures (0 unexpected) in 78.244 (78.252) seconds.
```

cmcgee1024:
@swift-ci test macOS

Sources/Swiftly/Uninstall.swift
justice-adams-apple: interesting, so by installing swiftly then you wouldn't really be able to invoke your original swift toolchain installation?

Say I had a toolchain installed, then I decide to install swiftly, I download some toolchains with swiftly, then ultimately uninstall them all. I then go to invoke swift again under the assumption my default toolchain which was installed prior to swiftly would take precedence. Sounds like in this case that would not work because all the symlinks would still be proxying through swiftly which wouldn't be able to find a toolchain?

cmcgee1024: Right, by installing swiftly you're giving it the ability to control which toolchain you are invoking. Note that since swiftly maintains its own record of toolchains it installed, any that you've installed previously wouldn't be known to swiftly since they could be in virtually any location on the system.

The user can take control temporarily by overriding their PATH to the external toolchain that they want to use, or uninstall swiftly.

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
@swift-ci test macOS

DESIGN.md
```diff
@@ -60,7 +67,7 @@ A simple setup for managing the toolchains could look like this:
 
 The toolchains (i.e. the contents of a given Swift download tarball) would be contained in the toolchains directory, each named according to the major/minor/patch version. `config.json` would contain any required metadata (e.g. the latest Swift version, which toolchain is selected, etc.). If pulling in Foundation to use `JSONEncoder`/`JSONDecoder` (or some other JSON tool) would be a problem, we could also use something simpler.
 
-The `~/.local/bin` directory would include symlinks pointing to the `bin` directory of the "active" toolchain, if any.
+The `~/.local/bin` directory would include symlinks pointing to swiftly itself. When the proxies binaries are executed swiftly proxies them to the requested toolchain, or the default.
```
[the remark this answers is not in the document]
cmcgee1024: Thanks, I've changed this to put the proxies and swiftly itself into its own managed directory by default, which is `~/.local/share/swiftly/bin`. This matches what we do on macOS, except there we use a more typical macOS location `~/Library/Application Support/swiftly/bin`.

DESIGN.md
```diff
@@ -266,6 +285,58 @@ This command checks to see if there are new versions of `swiftly` itself and upg
...
+# CMake
+swiftly run cmake -G ninja
+swiftly run ninja build
+
+# Autoconf
+swiftly run ./configure
+swiftly run make
+```
+
+Swiftly adjusts certain environment variables, such as prefixing the PATH to the selected toolchain directory, and setting the CC and CXX variables to the locations of clang and clang++ in those toolchains so that the build tools use them. If you want to explicitly specify a toolchain for the command you can do that with a selector notation like this:
```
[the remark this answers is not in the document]
cmcgee1024: I've removed the setting of `CC` and `CXX` variables, opting instead to have the user provide those if they need them, copied verbatim along with the rest of their environment. The only environment changes is to the `PATH` prepending the current toolchain, and leaving the rest of it untouched as the user had set previously.

Overriding the proxies in this case has the benefit of keeping this as a workaround for proxy performance issues. Also, this approach has `pyenv` as a precedent, which is documented to be doing the exact same thing. https://github.com/pyenv/pyenv/blob/master/COMMANDS.md#pyenv-exec

cmcgee1024:
@swift-ci test macOS

DESIGN.md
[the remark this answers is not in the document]
patrickfreed: `~/.local/bin` was originally chosen according to the XDG Base Directories specification, which is something of a standard for directories like this. While swiftly doesn't own `~/.local/bin`, it is the designated place to drop per-user executables according to that spec. Why would it be bad form to put our symlinks there?

https://specifications.freedesktop.org/basedir-spec/latest/
https://wiki.archlinux.org/title/XDG_Base_Directory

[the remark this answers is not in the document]
cmcgee1024: The concern is that the user might have swift toolchain, and other items (clang, lld, lldb, etc.) existing in the ~/.local/bin directory and swiftly will attempt to manage those.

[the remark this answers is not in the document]
al45tair: There are really _two_ reasons for this; the first is that, as @cmcgee1024 says, you might actually have Swift installed in `~/.local.bin`, in which case the proxies would then overwrite your Swift executables, which would be wrong. The second is that when using tools like `swiftly`, there are situations where you want to bypass the proxies, and it's easiest to do that if they're in a separate `bin` directory all of their own.

There is precedent here; `pyenv`, `rbenv` et al _also_ put their shims (which we seem to be calling proxies) into a separate `bin` directory for the same reasons.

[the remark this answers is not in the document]
al45tair: > Thanks, I've changed this to put the proxies **and swiftly itself** into its own managed directory by default

I think `swiftly` itself belongs in `~/.local/bin` or `/usr/local/bin` or wherever the user installed it. It's just the proxies that should live elsewhere IMO.

cmcgee1024:
@swift-ci test macOS

shahmishal:
@swift-ci test macOS

shahmishal:
@al45tair @justice-adams-apple When you get a change, can you review the latest changes?

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
@swift-ci test macOS

justice-adams-apple (approved):
LGTM

DESIGN.md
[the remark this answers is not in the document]
al45tair: I don't have a strong objection to that, I suppose, but it does mean that you can't use `swiftly` without having the proxies in your path if you've let `swiftly` install itself. At least, unless you make a symlink from `~/.local/bin` or similar.

al45tair (approved):
I haven't checked carefully through the code changes to `swiftly` itself, but I think the overall design seems fine. (I still have a slight reservation about the install location of the `swiftly` binary itself, but it's relatively minor and easily fixed by the user with a symlink if they care.)

PR #158, opened 2024-08-23
Integrate SwiftFormat into the swiftly project dependencies

cmcgee1024:
The SwiftFormat that is run in the CI is a very specific version that is unlikely to be the one a dev has installed with homebrew or discovered for themselves and built from source. Discovering the version involves finding the magic in the lint.dockerfile.

SwiftPM provides a mechanism for bringing in product dependencies so that you can simply run using the toolchain. The dependencies can be specified using an exact match on a particular version. These products from the dependencies can be discovered using the `swift package describe` command, and perhaps someday Swift devs will discover a package's dev tools in this way.

Add the SwiftFormat dependency as a package dependency in swiftly at the version that is currently used in the soundness CI checks. Use the the swift command to invoke the 'swiftformat' product from the dependency instead of manually git cloning and building it manually in the dockerfile.

Provide a rationale for the dependency in the Package.swift for anyone who is curious about it.

### DIFF
--- a/Package.resolved
+++ b/Package.resolved
@@ -1,4 +1,5 @@
 {
+  "originHash" : "942a1612e8d4b18479d16ed7518859be9ac852972e43afb2a08c65d1037a641f",
   "pins" : [
     {
       "identity" : "async-http-client",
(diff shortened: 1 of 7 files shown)


cmcgee1024:
@swift-ci test macOS

rauhul:
Given that swiftly is now part of Swiftlang, it would be good to adopt swift-format

cmcgee1024:
> Given that swiftly is now part of Swiftlang, it would be good to adopt swift-format

Yes, that makes sense. I've raised an issue to track that since there's more work involved than this PR can cover: #159

cmcgee1024:
@swift-ci test macOS

adam-fowler:
Is there any point doing this if the eventual plan is to move swift-format?

cmcgee1024:
> Is there any point doing this if the eventual plan is to move swift-format?

I think that moving to swift-format is something that we need to do, but it will take some effort.

An immediate problem that I hit is that I need to have the SwiftFormat repo checked out at the exact right version to run the formatter before I commit. With this change I can have certainty that when I run the formatter it will be in the exact same version and configuration as the soundness check in CI.

Also, consolidating the formatter into the project configuration is a stepping stone to switching the version of the formatter, and even switching formatters to swift-format.

adam-fowler (approved):

PR #161, opened 2024-08-26
Add installation of debian 12 and required dependencies

futurejones:
Debian 12 has been officially available since June 5th.

Nearly 3 months later `swiftly` still does not support this release.

I know there are moves to change how swiftly works with functions currently handled by the install script moved inside the `swiftly` executable binary.

The timeline for this is unknown and judging by the current level of draft PR's and comments could be many months away, if ever. 

Adding support for `macOS` and `Windows` should not be impacting on and delaying support for new `Linux` platforms.

I propose in the interim we add compatibility with the new supported platforms immediately in the current `swiftly-install` script.

This PR adds support for `Debian 12`, a similar approach can be used for the other new platforms.

### DIFF
--- a/install/swiftly-install.sh
+++ b/install/swiftly-install.sh
@@ -128,6 +128,18 @@ install_system_deps () {
         return
     fi
 
+    if [[ "$docker_platform_name" == "debian" ]]; then
+        echo "Detected Debian platform."
+        set +o errexit
+        if [[ "$(id --user)" == "0" ]]; then
+            apt-get install --quiet -y build-essential git gnupg2  libcurl4 libedit2 libncurses-dev libpython3-dev  libxml2  libz3-dev pkg-config python3 python3-lldb tzdata unzip zlib1g-dev
+        else
+            sudo apt-get install --quiet -y build-essential git gnupg2  libcurl4 libedit2 libncurses-dev libpython3-dev  libxml2  libz3-dev pkg-config python3 python3-lldb tzdata unzip zlib1g-dev
+        fi
+        return
+    fi
+
+
     dockerfile_url="https://raw.githubusercontent.com/apple/swift-docker/main/nightly-main/$docker_platform_name/$docker_platform_version/Dockerfile"
     dockerfile="$(curl --silent --retry 3 --location --fail $dockerfile_url)"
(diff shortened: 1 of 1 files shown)


install/swiftly-install.sh
```diff
@@ -128,6 +128,18 @@ install_system_deps () {
         return
     fi
 
+    if [[ "$docker_platform_name" == "debian" ]]; then
```
patrickfreed, line 131: FYI I filed https://github.com/swiftlang/swift-docker/issues/409 to cover adding the proper Debian 12 images so we don't need the hard code or fallback to swift-ci's list.

patrickfreed (commented):
Rather than merging support for just Debian 12, I think we'd want to resurrect #115, which covers the other newer platforms as well.

@cmcgee1024 I think it could be reasonable to push a quick update to swiftly-install.sh to support these new platforms while work progresses on the other bigger 0.4.0 work. WDYT?

install/swiftly-install.sh
futurejones: I think hard coding the dependencies is a much better option, especially for all the ubuntu and debian platforms. Trying to parse multiple dockerfiles to obtain the same dependency list is over engineering what should be a very simple task.
The following list of dependencies covers all current versions of ubuntu and debian with swift releases.
```bash
build-essential 
git 
gnupg2  
libcurl4 
libedit2 
libncurses-dev 
libpython3-dev  
libxml2  
libz3-dev 
pkg-config 
python3 
python3-lldb 
tzdata 
unzip 
zlib1g-dev
```
I'm sure a similar simplified list could be made for RHEL platforms.

If in some future swift release some extra dependencies are required, then it is a very quick and simple task to add it to the list.

Simple to update, simple to maintain and guaranteed to work correctly.

futurejones:
@patrickfreed @cmcgee1024 can we get some progress on this.

The reason I have restricted this PR to adding `debian 12` is so that can be easily processed, reviewed and merged.

It does not impact on, or effect any other platforms or distributions, and makes no difference to any future direction this project might want to make.

No new release of `swiftly` is needed so there is no reason why this can't get approved and merged.

futurejones:
closing this pull request due to lack of interest by the swiftly team in doing anything

PR #173, opened 2024-10-19
Add new Linux platforms: Ubuntu 24.04, Debian 12, Fedora 39

cmcgee1024:
Add support for these new platforms to swiftly, the autodetection, the tests, and docker support.

Remove the swiftly installer portion of the code.


cmcgee1024:
@swift-ci test macOS

cmcgee1024:
@swift-ci test macOS

Sources/LinuxPlatform/Linux.swift
```diff
@@ -481,98 +535,62 @@ public struct Linux: Platform {
...
-        var ubuntuCodeName: String?
         for info in releaseInfo.split(separator: "\n").map(String.init) {
             if info.hasPrefix("ID=") {
                 id = String(info.dropFirst("ID=".count)).replacingOccurrences(of: "\"", with: "")
             } else if info.hasPrefix("ID_LIKE=") {
                 idlike = String(info.dropFirst("ID_LIKE=".count)).replacingOccurrences(of: "\"", with: "")
             } else if info.hasPrefix("VERSION_ID=") {
-                versionID = String(info.dropFirst("VERSION_ID=".count)).replacingOccurrences(of: "\"", with: "")
-            } else if info.hasPrefix("UBUNTU_CODENAME=") {
-                ubuntuCodeName = String(info.dropFirst("UBUNTU_CODENAME=".count)).replacingOccurrences(of: "\"", with: "")
```
justice-adams-apple, line 520: Do we still need to account for stings which start with "UBUNTU_CODENAME" here?

Sources/LinuxPlatform/Linux.swift
```diff
@@ -481,98 +535,62 @@ public struct Linux: Platform {
...
                 } else {
                     print(message)
                 }
-                return self.manualSelectPlatform(platformPretty)
+                return await self.manualSelectPlatform(platformPretty)
             }
 
-            return PlatformDefinition(name: "ubi9", nameFull: "ubi9", namePretty: "RHEL 9")
+            return PlatformDefinition.rhel9
+        } else if let pd = [PlatformDefinition.ubuntu1804, .ubuntu2004, .ubuntu2204, .ubuntu2310, .ubuntu2404, .debian12, .fedora39].first(where: { $0.name == id + versionID }) {
```
justice-adams-apple, line 592: If we're gonna maintain a static data structure here, might as well make it a dictionary so we get better performance on look ups? Any thoughts? It's not really a big deal but since we're going to maintain the list of linux versions here might be worth it

Sources/LinuxPlatform/Linux.swift
```diff
@@ -125,6 +137,26 @@ public struct Linux: Platform {
                 "tzdata",
                 "zlib1g-dev",
             ]
+        case "ubuntu2404":
```
justice-adams-apple, line 140: I think we're missing 2310 from this list

Tests/SwiftlyTests/HTTPClientTests.swift
```diff
@@ -66,15 +77,17 @@ final class HTTPClientTests: SwiftlyTests {
                 // GIVEN: we have a swiftly http client with swift.org metadata capability
                 // WHEN: we ask for the first five releases of a supported platform in a supported arch
                 let releases = try await SwiftlyCore.httpClient.getReleaseToolchains(platform: platform, arch: arch, limit: 5)
-                // THEN: we get five releases
-                XCTAssertEqual(5, releases.count)
+                // THEN: we get at least 1 release
+                XCTAssertTrue(1 <= releases.count)
+
+                if newPlatforms.contains(platform) { continue } // Newer distros don't have main snapshots yet
```
justice-adams-apple, line 83: How will we eventually know to remove this for proper testing? Seems like the kind of thing I'd forget 😄

Tests/SwiftlyTests/SwiftlyTests.swift
```diff
@@ -99,25 +99,8 @@ class SwiftlyTests: XCTestCase {
...
 
     func baseTestConfig() async throws -> Config {
-        let getEnv = { varName in
-            guard let v = ProcessInfo.processInfo.environment[varName] else {
-                throw SwiftlyTestError(message: "environment variable \(varName) must be set in order to run tests")
-            }
-            return v
-        }
-
-        let name = try? getEnv("SWIFTLY_PLATFORM_NAME")
```
justice-adams-apple, line 109: were these variables just unused?

docker/docker-compose.2404.yaml
```diff
@@ -0,0 +1,24 @@
+version: "3"
+
+services:
+
+  test-setup:
```
justice-adams-apple, line 5: are there Dockerfiles for this?

install/swiftly-install.sh
```diff
@@ -1,716 +0,0 @@
-#!/usr/bin/env bash
```
justice-adams-apple, line 1: A bit behind I think, why did we remove all of these scripts + their tests?

justice-adams-apple (changes requested):

Sources/LinuxPlatform/Linux.swift
cmcgee1024: I couldn't find a list of packages for ubuntu 23.10 from the Dockerfiles. If someone knows where the docker file is, I can add it here.

This is where I usually look for them: https://github.com/swiftlang/swift-docker/tree/main/6.0/ubuntu

Tests/SwiftlyTests/HTTPClientTests.swift
cmcgee1024: Integration tests that are in the middle of the test pyramid like this a tricky, but necessary.

The trouble with mocking the responses is that we won't know if the API might change in some way that breaks swiftly. I think that we want to catch obvious integration problems between the swift.org API and what we can potentially request from it. If that means that we don't fully cover all possibilities with all supported platforms I think that's ok. This was a good test when I started working on the patch to verify that the platform definition strings were correct from the standpoint of the swift.org API.

Tests/SwiftlyTests/SwiftlyTests.swift
cmcgee1024: These variables have been slowly been superseded by platform auto detection, first in the production part of swiftly on macOS, then gradually the tests. This is removing the last remnant of them because I didn't want to miss test coverage of the auto-detection on the different CI platforms since I had modified some of that code here.

@shahmishal can we get the CI to add the new platforms? I think that I added all of the docker files, and docker compose, but the CI didn't pick up Debian 12, Fedora 39, or the two new Ubuntu (24.04, 23.10). I still see the usual RHEL9, Ubuntu 22.04, etc.

install/swiftly-install.sh
cmcgee1024: This bash script was the original `curl | bash` installation system for swiftly in the previous version. We've slowly migrated it to the new `swiftly init` mechanism for a few reasons, such as much better options handling, and a generally safer mechanism.

I'm removing it here because it's vestigial at this point, and confuses new contributors who still see this as the installer. Plus, it would need to have support for these new platforms.

Sources/LinuxPlatform/Linux.swift
```diff
@@ -408,59 +478,43 @@ public struct Linux: Platform {
...
-        case "2":
-            return PlatformDefinition.ubuntu2004
-        case "3":
-            return PlatformDefinition.ubuntu1804
-        case "4":
-            return PlatformDefinition.rhel9
-        case "5":
-            return PlatformDefinition.amazonlinux2
-        default:
+        guard let choiceNum = Int(choice) else {
```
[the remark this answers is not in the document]
cmcgee1024: I think that this guard will guard against non-integer input, such as "foo", or "bar". The guard below should check for 0, or something larger than the linuxPlatforms count and cancel the installation. The final array subscript at the return statement should be safely in range.

Sources/LinuxPlatform/Linux.swift
cmcgee1024: We're reading this file line-by-line and we don't have any need for the ubuntu codename anymore. I think it's safe to just skip ubuntu codename lines.

docker/docker-compose.2404.yaml
cmcgee1024: I believe that the docker files for these are covered by the images from docker hub, like `swift:5.10-noble`, and also the `test.dockerfile` and `install-test.dockerfile` which are ubuntu-based with some variables for the ubuntu version. These Dockerfiles should be invoked by the CI, but I don't know the magic hooks to make the CI try these new platforms now that we have the docker and docker compose files. Maybe @shahmishal can hook up CI for them?

Sources/LinuxPlatform/Linux.swift
cmcgee1024: I manually pieced together the list by looking at the differences between ubuntu 22.04, and 24.04. Tested it out in docker and it looks to be working fine. There's only the swift 5.10.1 release available for that one though, so I wonder how useful and used this version of ubuntu will be.

cmcgee1024:
@swift-ci test macOS

Sources/LinuxPlatform/Linux.swift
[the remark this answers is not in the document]
justice-adams-apple: oh I see, the guard clause tripped me up, Makes sense

docker/docker-compose.debian12.yaml
```diff
@@ -0,0 +1,24 @@
+version: "3"
+
+services:
+
+  test-setup:
+    image: swiftly:debian12-test
+    build:
```
justice-adams-apple, line 7: should this specify the newly added Dockerfile?

install/swiftly-install.sh
justice-adams-apple: But it was also responsible for pulling the latest release binary of swiftly right? So is the implication that going forward we should expect users to build from source? Or should we host release binaries on the repo? Just wondering if the readme/release process needs to be changed here

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
@swift-ci test macOS

install/swiftly-install.sh
cmcgee1024: There's a few possibilities:
* Direct download of the binary package from swift.org (tarball and macOS pkg) from the downloads page (see https://github.com/swiftlang/swift-org-website/pull/830 for the current draft)
* Reliable API or download URL where the package can be downloaded from swift.org
* System package managers (apt, yum, brew)

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
The macOS tests are all passing:

```
Test Suite 'All tests' passed at 2024-11-01 13:13:04.370.
	 Executed 66 tests, with 0 failures (0 unexpected) in 76.277 (76.285) seconds
```

install/swiftly-install.sh
0xTim: This should be done in a separate PR to discuss is rather than bundling it in as part of this PR

cmcgee1024:
@swift-ci test macOS

Sources/LinuxPlatform/Linux.swift
cmcgee1024: I have removed 23.10 from the list because we can't bring up infrastructure for a platform that is EOL.

docker/docker-compose.debian12.yaml
cmcgee1024: Since we've moved to the new GitHub Actions infrastructure I've removed the old docker compose files and Dockerfiles from here.

cmcgee1024:
We've moved to the new GitHub Actions infrastructure, so there's been a few changes here.

* Adjusted the .github/workflows/pull_request.yaml to add the three new Linux distributions so that the checks are run
* Adjusted some test cases to work across all of the new platforms given that some of them don't have 5.9 toolchains
* Skipped tests that rely on snapshots on platforms that don't currently have snapshots

cmcgee1024:
The macOS tests are passing:

```
Test Suite 'All tests' passed at 2024-11-08 13:27:54.065.
	 Executed 66 tests, with 0 failures (0 unexpected) in 77.773 (77.781) seconds
```

Package.resolved
```diff
@@ -150,8 +150,8 @@
       "kind" : "remoteSourceControl",
       "location" : "https://github.com/apple/swift-tools-support-core.git",
       "state" : {
-        "revision" : "3b13e439a341bbbfe0f710c7d1be37221745ef1a",
-        "version" : "0.6.1"
+        "revision" : "5b130e04cc939373c4713b91704b0c47ceb36170",
+        "version" : "0.7.1"
```
cmcgee1024, line 154: There was a compile error in the TSC code only when compiling in these new Linux platforms, likely due to a header change in libc. The new version fixes this problem.

Tests/SwiftlyTests/SwiftlyTests.swift
cmcgee1024: Switching over to GH actions has yielded these new platforms for testing. See the "checks" section of this PR.

docker/docker-compose.2404.yaml
cmcgee1024: In this latest version of the PR none of these are needed anymore since we switched to GH actions. The docker setup is managed by the official docker containers by the reusable workflow in `swiftlang/github-workflow`.

justice-adams-apple (approved):
LGTM

PR #178, opened 2024-10-21
Fix package install detection for packages that are partially installed

cmcgee1024:
On Debian/Ubuntu Linux that use the dpkg/apt package manager there is a possibility that a package is not in a fully installed state. It can be removed, leaving its configuration files, or the install could have failed. Check rigorously that the package is fully installed without error before producing the command-line.

Also, correct the install command to use apt-get instead of apt since that command-line is guaranteed to be stable.

### DIFF
--- a/Sources/LinuxPlatform/Linux.swift
+++ b/Sources/LinuxPlatform/Linux.swift
@@ -164,11 +164,11 @@ public struct Linux: Platform {
 
         let manager: String? = switch platformName {
         case "ubuntu1804":
-            "apt"
+            "apt-get"
         case "ubuntu2004":
-            "apt"
+            "apt-get"
         case "ubuntu2204":
-            "apt"
+            "apt-get"
         case "amazonlinux2":
             "yum"
         case "ubi9":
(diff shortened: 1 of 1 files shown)


cmcgee1024:
This experiment didn't fix the problem. Closing.

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
macOS tests are passing:

```
Test Suite 'All tests' passed at 2024-10-24 11:44:58.155.
	 Executed 66 tests, with 0 failures (0 unexpected) in 75.092 (75.100) seconds
```

Sources/LinuxPlatform/Linux.swift
```diff
@@ -216,21 +216,38 @@ public struct Linux: Platform {
...
+                    // The package might be listed but not in an installed non-error state.
+                    //
+                    // Look for something like this:
+                    //
+                    //   Desired=Unknown/Install/Remove/Purge/Hold
+                    //   | Status=Not/Inst/Conf-files/Unpacked/halF-conf/Half-inst/trig-aWait/Trig-pend
+                    //   |/ Err?=(none)/Reinst-required (Status,Err: uppercase=bad)
+                    //   ||/
+                    //   ii  pkgfoo         1.0.0ubuntu12        My description goes here....
+                    return pkgList.contains("\nii ")
```
justice-adams-apple, line 248: This works, but is there something we could look for which is more indicative of an error than `\nii`? 🤔 If not, LGTM

justice-adams-apple (approved):

Sources/LinuxPlatform/Linux.swift
cmcgee1024: If the `dpkg -l` command fails then that is treated as an indication that the package isn't installed, which throws an error caught below. It's just that this output can be that the package is partially installed, which requires this output check here. I don't think that there's much else we can do.

PR #181, opened 2024-10-29
Verify the libarchive payload

cmcgee1024:
The libarchive that is used to make a swiftly release for Linux is downloaded directly from the libarchive project on GitHub from a release artifact. In theory, this release could become tampered in the future.

Typically package managers get around this problem by both getting a specific version of the package, and also keep a hash/git commit to verify the contents.

Add a content check using an expected SHA-256 sum of the release source tarball of libarchive to help protect against any tampering of the release in the future.


cmcgee1024:
@swift-ci test macOS

justice-adams-apple (approved):

justice-adams-apple:
LGTM

PR #184, opened 2024-11-07
Add a github workflow for Linux based on the swiftlang template

cmcgee1024:
Test on the Linux distributions that Swiftly currently supports. Also, test with the latest stable Swift toolchain.

Build libarchive and install it using the shell script.


cmcgee1024:
@swift-ci test macOS

cmcgee1024:
The macOS tests are passing.

PR #188, opened 2024-11-23
Test swiftly release artifacts in a plain Linux environment

cmcgee1024:
Currently, there is no test coverage of a swiftly installation where there is no swift support at all, and nothing involving swiftly. Add a github workflow that performs a fresh swiftly install using the released artifact, installs the latest available swift toolchain, and verifies that the shell environment has been updated.

In order to facilitate this create an E2E test that performs these steps, but only in an environment where it can mutate the system. Also, make this test deployable to a plain Linux system with no swift support installed on it so that the tests can be run.


cmcgee1024:
@swift-ci test macOS

cmcgee1024:
The approach taken here for bundling the tests to run in a plain environment doesn't really work on macOS. Next step is probably to make a CLT that runs swiftly E2E tests on a system that it can mutate.

cmcgee1024:
Closing this in favour of #296

PR #191, opened 2024-12-04
Remove unnecessary C++ library package install from GH prep script

cmcgee1024:
These packages needed to be installed in some of the Linux distributions because the swift docker container images didn't include them, even though they are required to compile any C++ code. The docker images should be updated now and so these extra steps are no longer needed.


cmcgee1024:
#195 has this covered

PR #196, opened 2024-12-26
Adopt the static linux sdk

cmcgee1024:
With the Static Linux SDK swiftly can be compiled on a particular
Linux and processor architecture and the binary can run on a very
wide variety of versions and distributions. This eliminates the need
to build swiftly on "older" Linux distributions to produce a binary
that runs on the "newer" ones>

Bump the version of swift-tools-support-core so that it compiles
successfully in a Static SDK environment with Musl.

Fix the readLine function to add a newline to the prompt so that it
prints it out before waiting on user input on stdin.

Update the build-swiftly-release tool. First, detect the version of
swift so that the matching version of the Static SDK is downloaded
and installed into the toolchain. Use the Musl that is inside it to
configure and compile libarchive along with a special must-clang
wrapper script that sets the correct compile options to use Musl
instead of the system's glibc. Finally, build the swiftly binary with
the Static SDK / Musl linking it together with the Musl compiled
libarchive.

Deprecate the old flag that would decide between RHEL UBI9
or Amazon Linux 2 as the required release building platform and
eliminate the supported Linux check.

### DIFF
--- a/Sources/SwiftlyCore/SwiftlyCore.swift
+++ b/Sources/SwiftlyCore/SwiftlyCore.swift
@@ -45,7 +45,7 @@ public protocol InputProvider {
 public var inputProvider: (any InputProvider)?
 
 public func readLine(prompt: String) -> String? {
-    print(prompt, terminator: ": ")
+    print(prompt, terminator: ": \n")
     guard let provider = SwiftlyCore.inputProvider else {
         return Swift.readLine(strippingNewline: true)
     }
--- a/Tools/build-swiftly-release/BuildSwiftlyRelease.swift
+++ b/Tools/build-swiftly-release/BuildSwiftlyRelease.swift
@@ -371,8 +371,17 @@ struct BuildSwiftlyRelease: AsyncParsableCommand {
 
         FileManager.default.changeCurrentDirectoryPath(cwd)
 
-        // Statically link standard libraries for maximum portability of the swiftly binary
-        try runProgram(swift, "build", "--product=swiftly", "--pkg-config-path=\(pkgConfigPath)/lib/pkgconfig", "--static-swift-stdlib", "--configuration=release")
+        // Statically link standard libraries and use the static sdk for maximum portability
+#if arch(arm64)
+        let sdkName = "aarch64-swift-linux-musl"
+#else
+        let sdkName = "x86_64-swift-linux-musl"
+#endif
+
+        // FIXME: Adjust the URL and checksum to match the toolchain that is being used
+        try runProgram(swift, "sdk", "install" "https://download.swift.org/swift-6.0.3-release/static-sdk/swift-6.0.3-RELEASE/swift-6.0.3-RELEASE_static-linux-0.0.1.artifactbundle.tar.gz", "--checksum", "67f765e0030e661a7450f7e4877cfe008db4f57f177d5a08a6e26fd661cdd0bd")
+
+        try runProgram(swift, "build", "--swift-sdk", sdkName, "--product=swiftly", "--pkg-config-path=\(pkgConfigPath)/lib/pkgconfig", "--static-swift-stdlib", "--configuration=release")
 
         let releaseDir = cwd + "/.build/release"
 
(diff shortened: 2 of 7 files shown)


cmcgee1024:
@swift-ci test macOS

cmcgee1024:
@swift-ci test macOS

Sources/SwiftlyCore/SwiftlyCore.swift
```diff
@@ -45,7 +45,7 @@ public protocol InputProvider {
 public var inputProvider: (any InputProvider)?
 
 public func readLine(prompt: String) -> String? {
-    print(prompt, terminator: ": ")
+    print(prompt, terminator: ": \n")
```
justice-adams-apple, line 48: So does the user input then get output on a new line? So the stdout would look like:
```
Yes or No: 
No   <-- this is my user input
```
?

Tools/build-swiftly-release/BuildSwiftlyRelease.swift
```diff
@@ -13,6 +23,7 @@ public struct Error: LocalizedError {
 }
 
 public func runProgramEnv(_ args: String..., quiet: Bool = false, env: [String: String]?) throws {
+    print("\(args.joined(separator: " "))")
```
justice-adams-apple, line 26: do we want to print this if `quiet == False`?

(same question below)

Tools/build-swiftly-release/BuildSwiftlyRelease.swift
```diff
@@ -340,8 +293,38 @@ struct BuildSwiftlyRelease: AsyncParsableCommand {
...
+
+        guard let swiftRelease = swiftReleases.first(where: { ($0.name ?? "") == swiftVersion }) else {
+            throw Error(message: "Unable to find swift release using swift.org API: \(swiftVersion)")
+        }
+
+        guard let sdkPlatform = (swiftRelease.platforms ?? [SwiftPlatform]()).first(where: { ($0.name ?? "") == "Static SDK" }) else {
+            throw Error(message: "Swift release \(swiftVersion) has no Static SDK offering")
+        }
+
+        try runProgram(swift, "sdk", "install", "https://download.swift.org/swift-\(swiftVersion)-release/static-sdk/swift-\(swiftVersion)-RELEASE/swift-\(swiftVersion)-RELEASE_static-linux-0.0.1.artifactbundle.tar.gz", "--checksum", sdkPlatform.checksum ?? "deadbeef")
```
justice-adams-apple, line 323: the Static Linux SDK will always have this `.0.0.1` version identifier? Just want to make sure this doesn't break if that format changes

Tools/build-swiftly-release/BuildSwiftlyRelease.swift
```diff
@@ -371,8 +354,8 @@ struct BuildSwiftlyRelease: AsyncParsableCommand {
 
         FileManager.default.changeCurrentDirectoryPath(cwd)
 
-        // Statically link standard libraries for maximum portability of the swiftly binary
-        try runProgram(swift, "build", "--product=swiftly", "--pkg-config-path=\(pkgConfigPath)/lib/pkgconfig", "--static-swift-stdlib", "--configuration=release")
+        try runProgram(swift, "build", "--swift-sdk", "\(arch)-swift-linux-musl", "--product=swiftly", "--pkg-config-path=\(pkgConfigPath)/lib/pkgconfig", "--static-swift-stdlib", "--configuration=release")
+        try runProgram(swift, "sdk", "remove", sdkName)
```
justice-adams-apple, line 358: Could this inadvertently remove the linux SDK from a users machine even if they didn't want it to? As in they already had the static sdk installed, built swiftly, and now their SDK is gone and maybe they didn't expect that?

Tools/build-swiftly-release/BuildSwiftlyRelease.swift
cmcgee1024: It appears that this version is expected to change only very rarely, if ever.

justice-adams-apple (approved):

PR #205, opened 2025-01-25
Adopt OpenAPI for fetching current swiftly releases

cmcgee1024:
Use the OpenAPI specification for the swift.org API and the swift OpenAPI generation capabilities to avoid having to hand craft the HTTP handling and local data types.

### DIFF
--- a/Sources/SwiftlyCore/Error.swift
+++ b/Sources/SwiftlyCore/Error.swift
@@ -1,6 +1,6 @@
 import Foundation
 
-public struct Error: LocalizedError, CustomStringConvertible {
+public struct SwiftlyError: LocalizedError, CustomStringConvertible {
     public let message: String
 
     public init(message: String) {
--- a/Sources/SwiftlyCore/HTTPClient.swift
+++ b/Sources/SwiftlyCore/HTTPClient.swift
@@ -4,9 +4,12 @@ import Foundation
 import NIO
 import NIOFoundationCompat
 import NIOHTTP1
+import OpenAPIAsyncHTTPClient
+import OpenAPIRuntime
 
 public protocol HTTPRequestExecutor {
     func execute(_ request: HTTPClientRequest, timeout: TimeAmount) async throws -> HTTPClientResponse
+    func getCurrentSwiftlyRelease() async throws -> Components.Schemas.SwiftlyRelease
 }
 
 /// An `HTTPRequestExecutor` backed by the shared `HTTPClient`.
@@ -53,29 +56,25 @@ internal class HTTPRequestExecutorImpl: HTTPRequestExecutor {
     public func execute(_ request: HTTPClientRequest, timeout: TimeAmount) async throws -> HTTPClientResponse {
         try await self.httpClient.execute(request, timeout: timeout)
     }
+
+    public func getCurrentSwiftlyRelease() async throws -> Components.Schemas.SwiftlyRelease {
+        let config = AsyncHTTPClientTransport.Configuration(client: self.httpClient, timeout: .seconds(30))
+
+        // FIXME: use a middleware that adds the swiftly user agent header (problem is how to do this with async http client)
+        let client = Client(
+            serverURL: URL(string: "https://swift.org/api")!,
+            transport: AsyncHTTPClientTransport(configuration: config)
+        )
+
+        let response = try await client.getCurrentSwiftlyRelease()
+        return try response.ok.body.json
+    }
 }
 
 private func makeRequest(url: String) -> HTTPClientRequest {
     var request = HTTPClientRequest(url: url)
     request.headers.add(name: "User-Agent", value: "swiftly/\(SwiftlyCore.version)")
     return request
-}
-
-public enum SwiftOrgSwiftlyPlatformType: String, Codable {
-    case Darwin
-    case Linux
-    case Windows
-}
-
-public struct SwiftOrgSwiftlyPlatform: Codable {
-    public var platform: SwiftOrgSwiftlyPlatformType
-    public var x86_64: URL
-    public var arm64: URL
-}
-
-public struct SwiftOrgSwiftlyRelease: Codable {
-    public var version: SwiftlyVersion
-    public var platforms: [SwiftOrgSwiftlyPlatform]
 }
 
 struct SwiftOrgPlatform: Codable {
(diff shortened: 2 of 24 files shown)


cmcgee1024:
@swift-ci test macOS

Sources/SwiftlyCore/openapi-generator-config.yaml
```diff
@@ -0,0 +1,5 @@
+generate:
+  - types
+  - client
+namingStrategy: idiomatic
```
czechboy0, line 4: You can add a filter to only the exact operations you're using, to reduce the codegen and build times: https://swiftpackageindex.com/apple/swift-openapi-generator/1.7.0/documentation/swift-openapi-generator/configuring-the-generator#Document-filtering

Sources/SwiftlyCore/HTTPClient.swift
```diff
@@ -61,23 +77,6 @@ private func makeRequest(url: String) -> HTTPClientRequest {
...
-public struct SwiftOrgSwiftlyPlatform: Codable {
-    public var platform: SwiftOrgSwiftlyPlatformType
-    public var x86_64: URL
-    public var arm64: URL
-}
-
-public struct SwiftOrgSwiftlyRelease: Codable {
-    public var version: SwiftlyVersion
-    public var platforms: [SwiftOrgSwiftlyPlatform]
-}
```
czechboy0, line 79: You could keep these types, remove their Codable conformance, and add a eg `init(_ release: Components.Schemas.Release)` initialize on each of them that converts from the generated equivalent into this business logic type.

It's good practice to keep the API and business logic types separate, and explicitly convert between them. It then allows the API to keep evolving without it affecting your codebase too much (you'd only need to update the conversion initializer).

Sources/SwiftlyCore/HTTPClient.swift
```diff
@@ -53,6 +56,19 @@ internal class HTTPRequestExecutorImpl: HTTPRequestExecutor {
...
     }
+
+    public func getCurrentSwiftlyRelease() async throws -> Components.Schemas.SwiftlyRelease {
+        let config = AsyncHTTPClientTransport.Configuration(client: self.httpClient, timeout: .seconds(30))
+
+        // FIXME: use a middleware that adds the swiftly user agent header (problem is how to do this with async http client)
+        let client = Client(
+            serverURL: URL(string: "https://swift.org/api")!,
+            transport: AsyncHTTPClientTransport(configuration: config)
+        )
```
czechboy0, line 67: You can inject a user agent by adding the `middlewares` parameter on this Client. An example of a header injection middleware that you can copy is here: https://github.com/apple/swift-openapi-generator/blob/main/Examples/auth-client-middleware-example/Sources/AuthenticationClientMiddleware/AuthenticationClientMiddleware.swift

Sources/SwiftlyCore/Error.swift
```diff
@@ -1,6 +1,6 @@
 import Foundation
 
-public struct Error: LocalizedError, CustomStringConvertible {
+public struct SwiftlyError: LocalizedError, CustomStringConvertible {
```
czechboy0, line 3: I'd suggest you break out this renaming into a separate PR, it'll then make the OpenAPI changes a lot easier to review.

Sources/SwiftlyCore/HTTPClient.swift
cmcgee1024: Thanks for the suggestions. Swiftly doesn't do much with these types so they don't really add much value beyond holding the parts of the message payload that it uses right away. They don't go any further than swiftly itself. I was hoping to reduce the duplication and just use (or maybe extend) the ones generated by the OAG.

Sources/SwiftlyCore/Error.swift
cmcgee1024: Sorry for the noise. Thanks for having a look at this.

Sources/SwiftlyCore/HTTPClient.swift
cmcgee1024: Thanks, this example was what I needed to find the correct imports.

Sources/SwiftlyCore/openapi-generator-config.yaml
cmcgee1024: Thanks, I've applied a filter to the Swiftly portions of the API.

cmcgee1024:
@swift-ci test macOS

Sources/SwiftlyCore/HTTPClient.swift
```diff
@@ -53,6 +114,20 @@ internal class HTTPRequestExecutorImpl: HTTPRequestExecutor {
     public func execute(_ request: HTTPClientRequest, timeout: TimeAmount) async throws -> HTTPClientResponse {
         try await self.httpClient.execute(request, timeout: timeout)
     }
+
+    public func getCurrentSwiftlyRelease() async throws -> Components.Schemas.SwiftlyRelease {
+        let config = AsyncHTTPClientTransport.Configuration(client: self.httpClient, timeout: .seconds(30))
+        let swiftlyUserAgent = SwiftlyUserAgentMiddleware()
+
+        let client = Client(
+            serverURL: URL(string: "https://swift.org/api")!,
```
czechboy0, line 123: I think this either needs to be `https://www.swift.org/api/v1` (you're missing `/v1`), or just use the generated value: `serverURL: Servers.Server1.url()`.

Sources/SwiftlyCore/HTTPClient.swift
czechboy0: Yeah, the extensions you added also look good.

Sources/SwiftlyCore/HTTPClient.swift
cmcgee1024: Looks like a test gap. I will add an integration test case for this.

czechboy0 (approved):
Pending the Swift.org URL update, lgtm!

cmcgee1024:
> Pending the Swift.org URL update, lgtm!

Thanks again for your help with this. I think it should be ready for review and merge.

cmcgee1024:
@swift-ci test macOS

czechboy0 (approved):
Looks great!

cmcgee1024:
@swift-ci test macOS

Issue #223, opened 2025-02-07
Skipping Prompts when overwritting executables


johnbute:
Whenever you have existing executables and wish to install a version of swift, swiftly prompts the user to make sure that they want to overwrite those executables.

Installing package in user home directory...
The following existing executables will be overwritten:
...
Proceed? (y/N): 
Error: Toolchain installation has been cancelled

However, the program skips past the prompt and returns its default value (No), which cancels the installation.

cmcgee1024:
This should be fixed now. Thank you for raising this.

PR #226, opened 2025-02-11
Foreground process fix

johnbute:
Fixed a bug where a child process did not set main process back into foreground, causing issues such as not being able to let the user prompt y/n
Motivation:

When a child process is created and set into the foreground then terminates, the main process doesn't have the ability to prompt the user anymore.
Modifications:

Simply add a few lines to transfer the main process back to the foreground once the child process finishes

Result:
Allows for user interaction (user input), better resource management, and signal handling.

### DIFF
--- a/Sources/SwiftlyCore/Platform.swift
+++ b/Sources/SwiftlyCore/Platform.swift
@@ -211,6 +211,10 @@ extension Platform {
         }
         process.waitUntilExit()
 
+        if pgid != -1 {
+            tcsetpgrp(STDOUT_FILENO, pgid)
+        }
+
         guard process.terminationStatus == 0 else {
             throw RunProgramError(exitCode: process.terminationStatus, program: args.first!)
         }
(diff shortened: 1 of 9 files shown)


plemarquand:
Fixes #223

cmcgee1024:
@johnbute this is a really good fix to have. Thanks for tracking this down.

Can you reformat the code, commit and push it?

Sources/SwiftlyCore/Platform.swift
```diff
@@ -211,6 +211,10 @@ extension Platform {
         }
         process.waitUntilExit()
 
+        defer { if pgid != -1 {
```
cmcgee1024, line 214: suggestion: Move the defer just after the `if pgid != -1` block (line 211). This will put the reset right after the set code, which I think is a really good pattern visually in the code. Also, if something goes wrong in `process.waitUntilExit()` then this defer should reset the foreground process group in any case at the end of the scope.

Sources/SwiftlyCore/Platform.swift
```diff
@@ -254,6 +258,10 @@ extension Platform {
 
         process.waitUntilExit()
 
+        defer { if pgid != -1 {
```
cmcgee1024, line 261: suggestion: The same thing applies here. If this defer is moved before the process.waitUntilExit() then it will help with the locality of the action and the undo of that action. Also, it will work even if something goes wrong with the waitUntilExit() call.

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
@swift-ci test macOS

PR #238, opened 2025-02-28
Installation workflow improvements

cmcgee1024:
If the user happens to use the GUI installer to install the macOS pkg then
it will end up being install in /usr/local/bin/swift with no mechanism in place
for self-update. This patch redirects the GUI installer to install into the user
home directory.

For the easiest user experience swift.org may provide a copy-pastable set
of shell commands that can take the user directly to the `swiftly init` routine.
In this case there may be commands that can also take care of updating the
shell environment. In this case the init subcommand can take an option flag
to quiet the shell instructions.


kateinoigakukun:
I'm interested in the desired installation workflow you are thinking, so would you mind sharing the shape of the copy-pastable script? Are you planning to prohibit using the .pkg installer in GUI context?

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
@swift-ci test macOS

shahmishal (approved):

shahmishal:
@swift-ci test

shahmishal:
Is there a PR for `main`?

cmcgee1024:
@shahmishal it's here: https://github.com/swiftlang/swiftly/pull/249

PR #240, opened 2025-03-03
Make "swift-" prefix optional in MainSnapshotParser

kateinoigakukun:
Existing tools like `swiftenv` use `DEVELOPMENT-SNAPSHOT-YYYY-mm-dd-a` for main snapshots, without the `swift-` prefix. The `MainSnapshotParser` should be able to handle this and this change mirrors the `ReleaseSnapshotParser` in this regard, which already supports both `swift-` prefixed and unprefixed versions.

### DIFF
--- a/Sources/SwiftlyCore/ToolchainVersion.swift
+++ b/Sources/SwiftlyCore/ToolchainVersion.swift
@@ -411,17 +411,17 @@ struct ReleaseSnapshotParser: ToolchainSelectorParser {
     }
 }
 
-/// Parser for selectors like the following:
+/// Parser for selectors like the following (with optional "swift-" prefix):
 ///    - main-snapshot-YYYY-mm-dd
 ///    - main-snapshot
 ///    - main-SNAPSHOT-YYYY-mm-dd
 ///    - main-SNAPSHOT
-///    - swift-DEVELOPMENT-SNAPSHOT-YYYY-mm-dd-a
-///    - swift-DEVELOPMENT-SNAPSHOT-YYYY-mm-dd
-///    - swift-DEVELOPMENT-SNAPSHOT
+///    - DEVELOPMENT-SNAPSHOT-YYYY-mm-dd-a
+///    - DEVELOPMENT-SNAPSHOT-YYYY-mm-dd
+///    - DEVELOPMENT-SNAPSHOT
 struct MainSnapshotParser: ToolchainSelectorParser {
     static let regex: Regex<(Substring, Substring?)> =
-        try! Regex("^(?:main-snapshot|swift-DEVELOPMENT-SNAPSHOT|main-SNAPSHOT)(?:-([0-9]{4}-[0-9]{2}-[0-9]{2}))?(?:-a)?$")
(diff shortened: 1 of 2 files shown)


Sources/SwiftlyCore/ToolchainVersion.swift
```diff
@@ -411,17 +411,17 @@ struct ReleaseSnapshotParser: ToolchainSelectorParser {
     }
 }
 
-/// Parser for selectors like the following:
+/// Parser for selectors like the following (with optional "swift-" prefix):
 ///    - main-snapshot-YYYY-mm-dd
 ///    - main-snapshot
 ///    - main-SNAPSHOT-YYYY-mm-dd
 ///    - main-SNAPSHOT
-///    - swift-DEVELOPMENT-SNAPSHOT-YYYY-mm-dd-a
```
cmcgee1024, line 419: question: Since the swift prefix is optional with this change, should the examples with the swift prefix remain here to indicate that these are valid selectors?

Sources/SwiftlyCore/ToolchainVersion.swift
```diff
@@ -411,17 +411,17 @@ struct ReleaseSnapshotParser: ToolchainSelectorParser {
...
-///    - swift-DEVELOPMENT-SNAPSHOT-YYYY-mm-dd
-///    - swift-DEVELOPMENT-SNAPSHOT
+///    - DEVELOPMENT-SNAPSHOT-YYYY-mm-dd-a
+///    - DEVELOPMENT-SNAPSHOT-YYYY-mm-dd
+///    - DEVELOPMENT-SNAPSHOT
 struct MainSnapshotParser: ToolchainSelectorParser {
     static let regex: Regex<(Substring, Substring?)> =
-        try! Regex("^(?:main-snapshot|swift-DEVELOPMENT-SNAPSHOT|main-SNAPSHOT)(?:-([0-9]{4}-[0-9]{2}-[0-9]{2}))?(?:-a)?$")
+        try! Regex("^(?:swift-)?(?:main-snapshot|DEVELOPMENT-SNAPSHOT|main-SNAPSHOT)(?:-([0-9]{4}-[0-9]{2}-[0-9]{2}))?(?:-a)?$")
```
cmcgee1024, line 424: question: Does this regex now permit selections like `swift-main-snapshot-YYYY-mm-dd-a`? Does this make sense as a potential selector?

cmcgee1024 (commented):
Thank you for your contribution. Overall this looks good. I've added a couple of questions.

cmcgee1024:
@swift-ci test macOS

Sources/SwiftlyCore/ToolchainVersion.swift
kateinoigakukun: Yes, it's intentional. Given that we already accept `swift-a.b-snapshot-YYYY-mm-dd`, it makes `main` and release channel selectors more consistent.

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
Thanks again @kateinoigakukun for your contribution.

PR #244, opened 2025-03-07
Print file URLs without leading file schema

rauhul:
Swiftly interpolates some file URLs directly into user facing output leading to messages like 'The file `file:///Volumes/...` has been...' which needlessly contains the url schema prefix `file://`. This change updates Swiftly to use URL.path for better user output.


cmcgee1024 (approved):
Looks good to me, but I suspect that the formatting check will fail on line 131 of MacOS.swift.

cmcgee1024:
@swift-ci test macOS

PR #245, opened 2025-03-07
Make updates better and more resilient

cmcgee1024:


cmcgee1024:
@swift-ci test macOS

cmcgee1024:
TODO: update the main branch with these fixes

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
@swift-ci test macOS

Sources/Swiftly/Uninstall.swift
```diff
@@ -56,7 +56,12 @@ struct Uninstall: SwiftlyCommand {
             }
         } else {
             let selector = try ToolchainSelector(parsing: self.toolchain)
-            toolchains = startingConfig.listInstalledToolchains(selector: selector)
+            var installedToolchains = startingConfig.listInstalledToolchains(selector: selector)
+            // This is in the unusual case that the inUse toolchain is not listed in the installed toolchains
+            if let inUse = startingConfig.inUse, selector.matches(toolchain: inUse) && !startingConfig.installedToolchains.contains(inUse) {
+                installedToolchains.append(inUse)
```
justice-adams-apple, line 62: are we going to allow swiftly then to be able to uninstall toolchains it didn't have a hand in installing?

Sources/Swiftly/Install.swift
```diff
@@ -297,11 +297,18 @@ struct Install: SwiftlyCommand {
...
         // --use argument was provided.
-        if useInstalledToolchain || config.inUse == nil {
-            // TODO: consider adding the global default option to this commands flags
+        if useInstalledToolchain {
             try await Use.execute(version, globalDefault: false, &config)
         }
 
+        // We always update the global default toolchain if there is none set. This could
+        //  be the only toolchain that is installed, which makes it the only choice.
+        if config.inUse == nil {
```
justice-adams-apple, line 306: do we only want to do this when the user passes `--use` to install? In other words if `--use` was omitted does this still make sense?

Sources/Swiftly/Use.swift
```diff
@@ -233,5 +234,11 @@ public func selectToolchain(config: inout Config, globalDefault: Bool = false) a
         }
     }
 
-    return (config.inUse, .globalDefault)
+    // Check to ensure that the global default in use toolchain matches one of the installed toolchains, and return
+    // no selected toolchain if it doesn't.
+    guard let defaultInUse = config.inUse, config.installedToolchains.contains(defaultInUse) else {
+        return (nil, .globalDefault)
```
justice-adams-apple, line 240: Just to confirm: You shouldn't be able to select a toolchain that wasn't installed by swiftly?

Sources/Swiftly/Uninstall.swift
cmcgee1024: If the toolchain is in the `inUse` property of the `config.json` then it probably did have a hand in installing the toolchain at some point in time. This includes it in the list of installed toolchains for the purposes of the uninstall command. Below swiftly will expunge it from the config, and try once again to remove it from disk.

Sources/Swiftly/Install.swift
cmcgee1024: I think that it will be good to try and set a default global toolchain whenever we can to help users avoid cases where there's no default toolchain to fallback. It's an extra annoyance that most people probably don't care to have.

"Give me some kind of a swift toolchain please. I know that I have at least one installed in there."

Sources/Swiftly/Use.swift
cmcgee1024: This should be a rare case, caused by things like improper handling of the update operation that this PR should solve. Another potential case is race conditions. This is pure hardening of the code, and also giving the user a way back to a nominal state by `using` a toolchain that is in the list that swiftly installed.

justice-adams-apple (approved):
LGTM!

PR #246, opened 2025-03-08
Add support for Debian 12 in the install script

jakepetroules:
This makes the relatively mechanical changes needed to add support for Debian 12 in the install script. Debian Docker nightlies are not yet available, so we use Ubuntu's Dockerfile for computing the dependency list for now.

This has been tested on a Debian 12 installation and is able to correctly install swiftly.

### DIFF
--- a/swiftly-install.sh
+++ b/swiftly-install.sh
@@ -209,6 +209,24 @@ set_platform_rhel () {
     fi
 }
 
+set_platform_debian () {
+    # Debian Docker nightlies are not yet available, so use Ubuntu's Dockerfile for computing the dependency list for now
+
+    #docker_platform_name="debian"
+    docker_platform_name="ubuntu"
+    package_manager="apt-get"
+    export DEBIAN_FRONTEND=noninteractive
+
+    PLATFORM_NAME="debian$1"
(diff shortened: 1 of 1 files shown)


jakepetroules:
@swift-ci test

jakepetroules:
@cmcgee1024 Are you suggesting hardcoding the list for Debian for now, instead of falling back to the Ubuntu Dockerfile?

cmcgee1024:
@jakepetroules it would be consistent with the approach that we're taking with the latest in main/release 1.0. It will be good to have some kind of a API to get that information better than parsing Dockerfiles.

cmcgee1024:
@jakepetroules if you want to have Debian 12 support you could build your own swiftly from the main/release 1.0 branches and use it.

cmcgee1024:
This should be working with the latest release swiftly, with the exception of the ability to work with main-snapshots due to the missing metadata on swift.org.

kkebo:
> with the exception of the ability to work with main-snapshots due to the missing metadata on swift.org.

I submitted a PR to work with main-snapshots and 6.1-snaphsots.

https://github.com/swiftlang/swift-org-website/pull/948

PR #247, opened 2025-03-11
Fix typo in getting-started.md

shahmishal:


cmcgee1024:
@shahmishal if the user follows the macOS instructions then the swiftly binary should be in `~/usr/local/bin/swiftly`, and not in the system location `/usr/local/bin/swiftly`. I think that the instructions are correct as they are.

shahmishal:
I normally dont see binary getting install in `~/usr/local/bin` I have seen `/usr/local/bin`. 

Also noticed: 
```
~/usr/local/bin/swiftly init
Installing swiftly in /Users/user_name/Library/Application Support/swiftly/bin/swiftly...
```

So after running the command it moves to another location. Why not install the swiftly binary directly in `~/Library/Application Support/swiftly/bin`?

cmcgee1024:
> I normally dont see binary getting install in `~/usr/local/bin` I have seen `/usr/local/bin`.

There doesn't appear to be a standard for where binaries go in user's home directories that shells usually have in their default PATH. Linux is trying to set a standard for `/.local/bin`, but that doesn't seem to have gained much traction in the shells.

The original intent of picking `usr/local/bin` as the suffix in the macOS pkg was that if the user happened to install swiftly on the system under the root volume it would at least be added in `/usr/local/bin`, which is a very common place in most shells' default path. When the pkg is installed in the user's home directory it ends up being `~/usr/local/bin`. While this is not commonly on the default shell path, maybe it can be seen as an analog to a typical binary location.

We could change the pkg to install in `~/Library/Application Support/swiftly/bin` but then that's an even less likely place to be on a user's path before they install swiftly, and I expect that very few people will expect the swiftly binary to be in `/Library/Application Support/swiftly/bin` if the macOS pkg were to be installed on the system, which is not supported atm, but a possibility if the user decides to run the installer from the command-line and pick the root volume.

shahmishal:
My suggestion would be to install the binary in `~/Library/Application Support/swiftly/bin` and symlink from `~/.swiftly`  

For example: 

```
$HOME/.swiftly/swiftly -> $HOME/Library/Application Support/swiftly/bin
```

```
PATH=$HOME/.swiftly/swiftly:$PATH
```

Most of the applications installed for current user only create a directly in `$HOME`. For example, `~/.swiftpm` and `~/.rvm`

bnbarham:
> > I normally dont see binary getting install in `~/usr/local/bin` I have seen `/usr/local/bin`.
> 
> There doesn't appear to be a standard for where binaries go in user's home directories that shells usually have in their default PATH. Linux is trying to set a standard for `/.local/bin`, but that doesn't seem to have gained much traction in the shells.

While technically true (ie. there is no "standard" as such), from what I've seen the general go-to is either inside:
1. `.local`
2. `.<name>`

And then possibly a third on macOS:
3. `Library/Application Support/<name>`

FWIW my preference here would be (2), (3), (1).

> We could change the pkg to install in `Library/Application Support/swiftly/bin` but then that's an even less likely place to be on a user's path before they install swiftly

I'm not sure it's really any less likely than `~/usr/local/bin` - I've personally never seen this path used.

> and I expect that very few people will expect the swiftly binary to be in `/Library/Application Support/swiftly/bin`

Does this actually matter? If so, we could do both - actual binaries in `~/Library/Application Support/swiftly/bin` and symlinks from either `~/.local/bin` or `~/.swiftly/bin`. Possibly worth noting that SwiftPM does this today (though interestingly only into `~/Library` and not `~/Library/Application Support`):
```
> ls ~/.swiftpm
cache@ -> ~/Library/Caches/org.swift.swiftpm
configuration@ -> ~/Library/org.swift.swiftpm/configuration
security@ -> ~/Library/org.swift.swiftpm/security
swift-sdks@ -> ~/Library/org.swift.swiftpm/swift-sdks
```

PR #249, opened 2025-03-12
Installation workflow improvements

cmcgee1024:
Synthesize a product build and custom distribution plist.

Prepare for shell one-liner that update the shell environment.

Copy swiftly to the user home directory instead of install it in their home directory

Add more output to diagnose build release issues.


cmcgee1024:
@swift-ci test macOS

PR #258, opened 2025-03-26
updates the abstracts for each of the HowTo guides, and renames the topic heading from 'HOWTOS' to 'Guides'

heckj:
Updates the DocC content to make abstracts for each of the guide articles:
- renames the topic header from `HOWTOS` to `Guides`.
- sets an explicit abstract for each guide and makes them complete sentences to flow better on the top-level page.

### DIFF
--- a/Documentation/SwiftlyDocs.docc/automated-install.md
+++ b/Documentation/SwiftlyDocs.docc/automated-install.md
@@ -1,8 +1,9 @@
 # Install Swiftly Automatically
 
-Swiftly can be installed automatically in places like build/CI systems.
+Automatically install Swiftly.
 
-This guide will help you to script to the installation of swiftly and toolchains so that it can be unattended. We assume that you have working understanding of your build system. The examples are based on a typical Unix environment.
+This guide helps you to automate the installation of swiftly and toolchains so that it can run unattended, for example in build or continous integration systems.
+We assume that you have working understanding of your build system. The examples are based on a typical Unix environment.
 
 First, download the swiftly binary from swift.org for your operating system (e.g. Linux) and processor architecture (e.g. arm64, or x86_64). Here's an example using the popular curl command.
 
--- a/Documentation/SwiftlyDocs.docc/automated-install.md
+++ b/Documentation/SwiftlyDocs.docc/automated-install.md
@@ -1,6 +1,6 @@
 # Install Swiftly Automatically
 
-Automatically install Swiftly and Swift toolchains.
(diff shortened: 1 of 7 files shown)


Documentation/SwiftlyDocs.docc/automated-install.md
```diff
@@ -1,8 +1,9 @@
 # Install Swiftly Automatically
 
-Swiftly can be installed automatically in places like build/CI systems.
+Automatically install Swiftly.
```
cmcgee1024, line 3: suggestion: Let's use lowercase for swiftly here and below as the standard capitalization. The thinking is that it's very command-line centric, so that's the way that people interact with it for the most part. Also, this is to distinguish the term a bit more from Swift, because they can look very similar at a glance.

cmcgee1024:
@swift-ci test macOS

Documentation/SwiftlyDocs.docc/automated-install.md
heckj: In general, I get where you're coming from, but we're referencing swiftly inconsistently - capitalized references it as a proper noun - a formal name of what it is, and helps distinguish it from the adverb that it's overloading. What would you think instead of `Automatically install Swiftly and Swift toolchains.`

Documentation/SwiftlyDocs.docc/uninstall-toolchains.md
```diff
@@ -1,6 +1,6 @@
 # Uninstall Swift Toolchains
 
-swiftly uninstall
+Uninstall swift toolchains.
```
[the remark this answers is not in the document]
heckj: mea culpa - this should have been capitalized up front.

Documentation/SwiftlyDocs.docc/automated-install.md
patrickfreed: I want to echo @cmcgee1024's point here: the lowercase "swiftly" spelling is intentional to emphasize its identity as a CLI tool. Note that in the README swiftly is consistently referred to with this lowercase form. In practice needing to disambiguate with the regular "swiftly" adverb usage will not be a concern for our purposes, but seeing the capital S when referring to a CLI tool will in practice be slightly jarring to readers (I know it is to me, at least).

heckj: Got it - your docs, ultimately your call, so I'll go with it.

cmcgee1024:
Thank you for your contribution @heckj

PR #259, opened 2025-03-27
Change the default install path for macOS pkgs

cmcgee1024:
Correct productbuild issue when creating a pkg from synthesized
distribution.xml.

Fix quiet option so that it doesn't prevent the environment file from
being added to the profile.

### DIFF
--- a/Sources/Swiftly/Init.swift
+++ b/Sources/Swiftly/Init.swift
@@ -89,22 +89,6 @@ internal struct Init: SwiftlyCommand {
             }
         }
 
-        // Ensure swiftly doesn't overwrite any existing executables without getting confirmation first.
-        let swiftlyBinDir = Swiftly.currentPlatform.swiftlyBinDir
-        let swiftlyBinDirContents = (try? FileManager.default.contentsOfDirectory(atPath: swiftlyBinDir.path)) ?? [String]()
-        let willBeOverwritten = Set(["swiftly"]).intersection(swiftlyBinDirContents)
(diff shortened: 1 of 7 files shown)


cmcgee1024:
@swift-ci test macOS

Sources/Swiftly/Init.swift
```diff
@@ -89,22 +89,6 @@ internal struct Init: SwiftlyCommand {
             }
         }
 
-        // Ensure swiftly doesn't overwrite any existing executables without getting confirmation first.
```
cmcgee1024, line 92: This check had existed to check only if swiftly was already at the destination, prompting before overwriting it. The init subcommand is there to install swiftly, so it's not technically a problem if it's overwriting the binary in the destination. There is a separate check to see if it will be overwriting a user's configuration, and/or toolchains.

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
@swift-ci test macOS

justice-adams-apple (approved):
LGTM, could this change result in a user having two swiftly on path then (if they already had it installed at the previous `Library/Application Support` location)?

cmcgee1024:
> LGTM, could this change result in a user having two swiftly on path then (if they already had it installed at the previous Library/Application Support location)?

Yes, but we haven't had an official release of swiftly yet with macOS, so it should be manageable. There is always a general risk that the user could have multiple installs of swiftly in different locations, and we might not be able to easily detect that. One of them will ultimately win in the sense that the env vars override the others at the shell.

PR #263, opened 2025-03-27
Change the default install path for macOS pkgs (#259)

cmcgee1024:
Correct productbuild issue when creating a pkg

Set the default swiftly home and binary locations on macOS to ~/.swiftly


cmcgee1024:
@swift-ci test macOS

PR #264, opened 2025-03-28
Update the abstracts for the HowTo guides, Rename HOWTOS to Guides

cmcgee1024:
Updates the abstracts for each of the HowTo guides, and renames the topic heading from 'HOWTOS' to 'Guides'.


cmcgee1024:
@swift-ci test macOS

PR #273, opened 2025-03-29
Polish welcome message

SimplyDanny:
This applies some polishing on the textual output of the command line tool, namely:

* Inline the `hash -r` advice.
* Mention the location where toolchains go.
* Improve text a little generally.
* Restructure the code to built up the message.

### DIFF
--- a/Sources/Swiftly/Init.swift
+++ b/Sources/Swiftly/Init.swift
@@ -86,16 +86,27 @@ struct Init: SwiftlyCommand {
             if !skipInstall {
                 msg += """
 
-                    Once swiftly is set up, it will install the latest available Swift toolchain.
-
+                    Once swiftly is set up, it will install the latest available Swift toolchain. This can be
+                    suppressed with the '--skip-install' option.
                     """
                 #if os(Linux)
                 msg += """
-                    In the process, swiftly will add swift.org GnuPG keys into your keychain to verify
-                    the integrity of the downloads.
+                     In the process, swiftly will add swift.org
+                    GnuPG keys into your keychain to verify the integrity of the downloads.
 
                     """
+                #else
+                msg += "\n"
                 #endif
+            }
+            if !noModifyProfile {
+                msg += """
+
+                    For your convenience, swiftly will also attempt to modify your shell's profile file to make
+                    installed items available in your environment upon login. This can be suppressed with the
+                    '--no-modify-profile' option.
+
+                    """
             }
 
             SwiftlyCore.print(msg)
--- a/Sources/Swiftly/Init.swift
+++ b/Sources/Swiftly/Init.swift
@@ -67,6 +67,12 @@ struct Init: SwiftlyCommand {
         // Give the user the prompt and the choice to abort at this point.
         if !assumeYes {
             var msg = """
+                Welcome to swiftly, the Swift toolchain manager for Linux and macOS!
+
+                Please read the following information carefully before proceeding with the installation. If you
+                wish to customize the steps performed during the installation process, refer to 'swiftly init -h'
+                for configuration options.
+
                 Swiftly installs files into the following locations:
 
                 \(Swiftly.currentPlatform.swiftlyHomeDir.path) - Directory for configuration files
--- a/Sources/Swiftly/Init.swift
+++ b/Sources/Swiftly/Init.swift
@@ -66,24 +66,33 @@ struct Init: SwiftlyCommand {
 
         // Give the user the prompt and the choice to abort at this point.
         if !assumeYes {
-#if os(Linux)
-            let sigMsg = " In the process of installing the new toolchain, swiftly will add swift.org GnuPG keys into your keychain to verify the integrity of the downloads."
-#else
-            let sigMsg = ""
-#endif
-            let installMsg = if !skipInstall {
-                "\nOnce swiftly is installed it will install the latest available Swift toolchain.\(sigMsg)\n"
-            } else { "" }
+            var msg = """
+                Swiftly installs files into the following locations:
 
-            SwiftlyCore.print("""
-            Swiftly will install files into the following locations:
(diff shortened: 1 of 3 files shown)


Sources/Swiftly/Init.swift
```diff
@@ -67,21 +67,21 @@ struct Init: SwiftlyCommand {
...
+                "\nOnce swiftly is installed it will install the latest available Swift toolchain.\(sigMsg)\n"
             } else { "" }
 
             SwiftlyCore.print("""
-            Swiftly will be installed into the following locations:
+            Swiftly will install files into the following locations:
 
-            \(Swiftly.currentPlatform.swiftlyHomeDir.path) - Data and configuration files directory including toolchains
-            \(Swiftly.currentPlatform.swiftlyBinDir.path) - Executables installation directory
+            \(Swiftly.currentPlatform.swiftlyHomeDir.path) - Directory for configuration files
```
cmcgee1024, line 81: suggestion (blocking): Toolchains will also go into the swiftly home directory, but only on Linux. Putting a platform-specific description should help and clarify what's going where, especially really big things like the toolchains.

cmcgee1024 (commented):
praise: Thank you for making the intro screen much clearer and cleaner.

I just have one suggestion, and then I think this is good to be merged.

Sources/Swiftly/Init.swift
```diff
@@ -67,21 +67,21 @@ struct Init: SwiftlyCommand {
...
             let sigMsg = ""
 #endif
             let installMsg = if !skipInstall {
-                "\nOnce swiftly is installed it will install the latest available swift toolchain.\(sigMsg)\n"
+                "\nOnce swiftly is installed it will install the latest available Swift toolchain.\(sigMsg)\n"
             } else { "" }
 
             SwiftlyCore.print("""
-            Swiftly will be installed into the following locations:
+            Swiftly will install files into the following locations:
```
heckj, line 79:
```suggestion
            Swiftly installs files into the following locations:
```
(removes the future tense, which makes the action more direct)

heckj (commented):
Good updates - I'd suggest a minor additional tweak in one place

Sources/Swiftly/Init.swift
SimplyDanny: While it's okay to mention the toolchain location only on Linux, I felt that it's also a useful information on macOS.

Let me know if you have a strong objection against that or other ideas for improvement.

Sources/Swiftly/Install.swift
```diff
@@ -112,10 +112,8 @@ struct Install: SwiftlyCommand {
         // Fish doesn't cache its path, so this instruction is not necessary.
         if pathChanged && !shell.hasSuffix("fish") {
             SwiftlyCore.print("""
-            NOTE: We have updated some elements in your path and your shell may not yet be
-            aware of the changes. You can run this command to update your shell.
-
-                hash -r
+            NOTE: Swiftly has updated some elements in your path and your shell may not yet be
+            aware of the changes. You can run 'hash -r' to update your shell in place.
```
cmcgee1024, line 116: question: Can the command go onto its own line here? While the indentation is not very consistent at the moment, the idea is to have each command the user is being asked to run on separate lines so that they can select, copy, paste, and run up them more easily than trying to select from inside a paragraph.

Sources/Swiftly/Update.swift
```diff
@@ -137,10 +137,8 @@ struct Update: SwiftlyCommand {
 
         if pathChanged {
             SwiftlyCore.print("""
-            NOTE: We have updated some elements in your path and your shell may not yet be
-            aware of the changes. You can run this command to update your shell.
-
-                hash -r
+            NOTE: Swiftly has updated some elements in your path and your shell may not yet be
+            aware of the changes. You can run 'hash -r' to update your shell.
```
cmcgee1024, line 141: Same question as above.

Sources/Swiftly/Init.swift
cmcgee1024: This is good, thank you. I just didn't want to leave out any mention of where toolchains are going since they're such a vital part of swiftly and its workflows.

Sources/Swiftly/Install.swift
SimplyDanny: I've made it so that the command is on a separate line easy to copy and added the info that restarting the shell is an option, too. I personally like it when the reading flow isn't interrupted by colons.

Sources/Swiftly/Init.swift
```diff
@@ -66,24 +66,33 @@ struct Init: SwiftlyCommand {
...
+                    Once swiftly is set up, it will install the latest available Swift toolchain.
+
+                    """
+                #if os(Linux)
+                msg += """
+                    In the process, swiftly will add swift.org GnuPG keys into your keychain to verify
+                    the integrity of the downloads.
+
+                    """
+                #endif
```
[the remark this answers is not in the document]
SimplyDanny: That's done now. The logic is as before: When suppressing commands are not given, the text is printed to let the user know about this significant change. It now also mentions the command line option that'd suppress it right away.

SimplyDanny:
The more I added and customized the output, the more convinced I became that swiftly would benefit from adopting [Noora](https://noora.tuist.dev). It was made for exactly the use case of guiding users through steps and processes on the command line. Just in case you didn't know about Noora as of yet ... 😉

cmcgee1024:
@SimplyDanny this looks really good. It's currently failing the formatting check. Can you run the following to fix the formatting, and update the PR?

```
swift run swiftformat .
```

SimplyDanny:
> @SimplyDanny this looks really good. It's currently failing the formatting check. Can you run the following to fix the formatting, and update the PR?
> 
> ```
> swift run swiftformat .
> ```

That's done.

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
Thank you @SimplyDanny for your contribution.

PR #285, opened 2025-03-31
Fix typo in `DESIGN.md`

MaxDesiatov:
There's no such product or entity as "SPM", this contraction is just as invalid as "MOS" would be for macOS, or "SNIO" for SwiftNIO. All marketing and documentation materials consistently refer to it as "SwiftPM", and we should stay consistent in our codebase too. This also simplifies code search, as anything named as "SPM" almost never shows up in results when searching globally for "SwiftPM".

### DIFF
--- a/DESIGN.md
+++ b/DESIGN.md
@@ -341,7 +341,7 @@ CC=clang swiftly run +latest make
 
 ## Detailed Design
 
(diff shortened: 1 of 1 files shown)


cmcgee1024:
This looks good. Thank you @MaxDesiatov

PR #299, opened 2025-04-03
Remove some dead code and deduplicate some code in the HTTP request handling

jakepetroules:
Some of this was no longer used after #253


cmcgee1024:
I've incorporated these changes into a larger reorganization and cleanup activity in #304

PR #309, opened 2025-04-07
Convert file downloading to go through the OpenAPI abstraction

jakepetroules:
This is a little awkward due to the OpenAPI spec not supporting variable length path components in an endpoint declaration (that's coming in v4 of the spec), but it does the job to get all of our HTTP traffic flowing through the same abstraction layer.


Sources/SwiftlyCore/openapi.yaml
```diff
@@ -4,19 +4,163 @@ info:
   description: API for retrieving Swift version release information, including date, platform support, tags, and toolchain details.
   version: 1.1.0
 servers:
-  - url: https://www.swift.org/api/v1
+  - url: https://www.swift.org
     description: The production deployment.
-  - url: http://127.0.0.1:4000/api/v1
+  - url: https://download.swift.org
```
cmcgee1024, line 9: suggestion: the swift-org-website already has an OpenAPI specification for the download server in a separate document. Can the download-specific portions of this be pulled out into a replica of that document here? It would make it easier to update these ones when occasionally to pull in newer revisions. Also, the paths can be kept as they were originally without the "/api/v1" prefix.

For reference the two spec files are here for the website:
https://github.com/swiftlang/swift-org-website/tree/main/openapi

cmcgee1024:
@swift-ci test macOS

Sources/SwiftlyCore/openapi.yaml
jakepetroules: Done!

PR #310, opened 2025-04-08
Migrate swiftly to Swift 6 with concurrency checks

cmcgee1024:


cmcgee1024:
@swift-ci test macOS

Tests/SwiftlyTests/SwiftlyTests.swift
```diff
@@ -423,7 +423,7 @@ public enum SwiftlyTests {
     }
 }
 
-public class TestOutputHandler: SwiftlyCore.OutputHandler {
+public actor TestOutputHandler: SwiftlyCore.OutputHandler {
     public var lines: [String]
```
plemarquand, line 427: nit:
```suggestion
    public private(set) var lines: [String]
```

Sources/SwiftlyCore/ToolchainVersion.swift
```diff
@@ -399,11 +403,12 @@ struct StableReleaseParser: ToolchainSelectorParser {
 ///    - swift-a.b-DEVELOPMENT-SNAPSHOT-YYYY-mm-dd
 ///    - swift-a.b-DEVELOPMENT-SNAPSHOT
 struct ReleaseSnapshotParser: ToolchainSelectorParser {
-    static let regex: Regex<(Substring, Substring, Substring, Substring?)> =
+    static func regex() -> Regex<(Substring, Substring, Substring, Substring?)> {
```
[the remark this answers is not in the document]
cmcgee1024: No, it's because Regex is not Sendable in its current state. A lot of swift code just constructs regexes where they are used, so this would be no different. Maybe someday Regex can be Sendable.

Tests/SwiftlyTests/SwiftlyTests.swift
cmcgee1024: Fixed, thanks.

Sources/SwiftlyCore/ToolchainVersion.swift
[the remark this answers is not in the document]
cmcgee1024: There's a forum thread on this topic:

https://forums.swift.org/t/should-regex-be-sendable/69529/15

Tests/SwiftlyTests/SwiftlyTests.swift
cmcgee1024: Done

plemarquand (approved):

PR #313, opened 2025-04-09
Add version check and prompt for self update

### LINKED ISSUE #233: Gentle reminder about updated swiftly
Since swiftly is often making round-trips with the swift.org API for various commands, it could also do a brief check to see if it is up-to-date at the same time. If it discovers that it is out-of-date then there can be a warning message prompting someone to run `self-update`.

louisunlimited:
- Fixes #233 

Manually added checks at the start of each command (apart from `self-update` and `init`) to check for version mismatches, if found, it will try to prompt the user with:
```
"A new release of swiftly is available"
"Please run `swiftly self-update` to update."
```

It also tries to prompt the user at the very last stage of the command, ensuring the update prompt will always show at the end to avoid being buried.


louisunlimited:
I'm thinking about possibly refactoring this in some way to make things coherent, maybe all as a method under `SwiftlyCommand` like the existing:
https://github.com/swiftlang/swiftly/blob/2cedc32495c1e80faf84321ff2d1a4c6b3b82d30/Sources/Swiftly/Swiftly.swift#L83

Sources/Swiftly/Install.swift
```diff
@@ -79,6 +79,15 @@ struct Install: SwiftlyCommand {
     mutating func run(_ ctx: SwiftlyCoreContext) async throws {
         try validateSwiftly(ctx)
 
+        let swiftlyRelease = try await ctx.httpClient.getCurrentSwiftlyRelease()
```
cmcgee1024, line 82: question: can we put this code in validateSwiftly() as a kind of validation check? I think that it might make sense to have it there.

Sources/Swiftly/Install.swift
```diff
@@ -79,6 +79,15 @@ struct Install: SwiftlyCommand {
...
+        let shouldUpdateSwiftly = try swiftlyRelease.swiftlyVersion > SwiftlyCore.version
+        defer {
+            if shouldUpdateSwiftly {
+                ctx.print("A new release of swiftly is available")
```
cmcgee1024, line 86: suggestion: Print these messages to stderr so that it doesn't interfere with other tools that might be parsing stdout. It also has the benefit of being able to quieted at the shell if a user wants using redirects like this: `swiftly install 2>/dev/null`

Sources/Swiftly/Install.swift
cmcgee1024: I see that one problem is with the defer block. Perhaps the validate function can return a function that should be deferred by the caller?

cmcgee1024 (commented):
praise: this is a great start on this feature!

Sources/Swiftly/Install.swift
louisunlimited: Is there a predefined way of doing prints to stderr? Currently I'm using fileHandler to route to stderr, I see that print() depends on the context and would it be better modifying outputHandler in the context to be stderr and then change it back to being stdout?

cmcgee1024: I think that it's ok in this context to print directly to stderr outside of the ctx.print(). The context is there for the purpose of mocks during testing, and I don't think that this feature warrants testing of what it prints to stderr.

In terms of how to print to stderr, I think that these FileHandle API's in Foundation could work for this:
https://developer.apple.com/documentation/foundation/filehandle/standarderror
https://developer.apple.com/documentation/foundation/filehandle/write(contentsof:)

louisunlimited: Oh nice, that's exactly what I used for printing out to stderr, I'll mark this pr for review then, thanks!

cmcgee1024:
It looks like the big URL -> FilePath created conflicts on this PR.

louisunlimited:
Hi @cmcgee1024 I just merged main into this branch, lmk if you need anything else!

Sources/Swiftly/Swiftly.swift
```diff
@@ -107,5 +107,23 @@ extension SwiftlyCommand {
...
         _ = try await Config.load(ctx)
+
+        let swiftlyRelease = try await ctx.httpClient.getCurrentSwiftlyRelease()
+        let shouldUpdateSwiftly = try swiftlyRelease.swiftlyVersion > SwiftlyCore.version
+        
+        return {
+            if shouldUpdateSwiftly {
+                let updateMessage = """
+                -----------------------------
+                A new release of swiftly is available
```
cmcgee1024, line 118: nitpick: This sentence needs a period at the end.

cmcgee1024:
@swift-ci test macOS

Sources/Swiftly/Swiftly.swift
```diff
@@ -107,5 +107,23 @@ extension SwiftlyCommand {
 
         // Verify that the configuration exists and can be loaded
         _ = try await Config.load(ctx)
+
+        let swiftlyRelease = try await ctx.httpClient.getCurrentSwiftlyRelease()
```
cmcgee1024, line 111: issue(blocking): This will throw an error if there's some kind of an HTTP connection problem fetching the current swiftly release. That shouldn't abort the principal operation, such as use/install/etc.

suggestion: Change this to `try? await` so that the release is optional. Only if the release can be found and it is larger than the current version would trigger the message.

cmcgee1024:
@louisunlimited I suspect that the tests are failing because mocked http executor is producing fatal errors due to this attempt to make a network request. The idea is to keep the vast majority of tests from making real network requests so that they can run quickly and are immune to changes on swift.org that make them brittle.

Perhaps this could produce a mocked swiftly version instead, based on the current version in SwiftlyVersion?

cmcgee1024:
@louisunlimited you can fix the formatting error by running `swift run swiftformat .` and then commit the formatting fixes.

Sources/Swiftly/Swiftly.swift
louisunlimited: Sounds good, done

louisunlimited:
> @louisunlimited I suspect that the tests are failing because mocked http executor is producing fatal errors due to this attempt to make a network request. The idea is to keep the vast majority of tests from making real network requests so that they can run quickly and are immune to changes on swift.org that make them brittle.
> 
> Perhaps this could produce a mocked swiftly version instead, based on the current version in SwiftlyVersion?

Yeah I was going through the local tests and I see that by wrapping tests inside something like
```swift
try await SwiftlyTests.withMockedSwiftlyVersion(latestSwiftlyVersion: SwiftlyVersion(major: SwiftlyCore.version.major, minor: 0, patch: 0)) {
    ...
}
``` 
would fix it. I could just wrap this around all the tests (or the helper func they call), but was thinking if there's a better way. Any suggestions?

cmcgee1024:
> would fix it. I could just wrap this around all the tests (or the helper func they call), but was thinking if there are better ways. Any suggestions?

I think that's pretty consistent with what's being done in other cases in the test suite, so that seems fine to me.

louisunlimited:
@cmcgee1024 Just updated the tests! Lmk what you think, and also if it's necessary to add a "shouldPromptVersionUpate" test for each of the commands

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
Thanks for your contribution @louisunlimited

PR #315, opened 2025-04-09
Add ability to temporarily unlink swiftly

plemarquand:
Adds two new commands, `swiftly unlink` and `swiftly link`, which will disable and reenable swiftly's management of the active toolchain.

The `unlink` command removes the symlinks to toolchain binaries in the swiftly bin directory that is in the user's path. This allows the rest of the `$PATH` to be searched for available toolchain installations, falling back to the system default. On macOS with Xcode installed this has the effect of falling back to the toolchain in the user's installed Xcode.

The `link` command reinstates the symlinks to the `inUse` toolchain, which allows swiftly to resume management of the active toolchain.

### DIFF
--- a/Sources/Swiftly/Install.swift
+++ b/Sources/Swiftly/Install.swift
@@ -199,7 +199,8 @@ struct Install: SwiftlyCommand {
             selector = try ToolchainSelector(parsing: version)
         } else {
             if case let (_, result) = try await selectToolchain(ctx, config: &config),
-               case let .swiftVersionFile(_, sel, error) = result {
+               case let .swiftVersionFile(_, sel, error) = result
+            {
                 if let sel = sel {
                     selector = sel
                 } else if let error = error {
--- a/Sources/Swiftly/Install.swift
+++ b/Sources/Swiftly/Install.swift
@@ -86,6 +86,7 @@ struct Install: SwiftlyCommand {
         defer {
             versionUpdateReminder()
         }
+        try await valitateLinked(ctx)
 
         var config = try await Config.load(ctx)
         let toolchainVersion = try await Self.determineToolchainVersion(ctx, version: self.version, config: &config)
--- a/Sources/Swiftly/Install.swift
+++ b/Sources/Swiftly/Install.swift
@@ -89,30 +89,6 @@ struct Install: SwiftlyCommand {
 
         var config = try await Config.load(ctx)
 
-        var selector: ToolchainSelector
-
-        if let version = self.version {
-            selector = try ToolchainSelector(parsing: version)
-        } else {
-            if case let (_, result) = try await selectToolchain(ctx, config: &config),
-               case let .swiftVersionFile(_, sel, error) = result
-            {
-                if let sel = sel {
-                    selector = sel
-                } else if let error = error {
-                    throw error
-                } else {
-                    throw SwiftlyError(message: "Internal error selecting toolchain to install.")
-                }
-            } else {
-                throw SwiftlyError(
-                    message:
-                    "Swiftly couldn't determine the toolchain version to install. Please set a version like this and try again: `swiftly install latest`"
-                )
-            }
-        }
-
-        let toolchainVersion = try await Self.resolve(ctx, config: config, selector: selector)
         let (postInstallScript, pathChanged) = try await Self.execute(
             ctx,
             version: toolchainVersion,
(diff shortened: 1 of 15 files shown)


Sources/Swiftly/Install.swift
```diff
@@ -161,6 +138,101 @@ struct Install: SwiftlyCommand {
         }
     }
 
+    public static func setupProxies(
```
cmcgee1024, line 141: thought: I think that if the setup proxies has no side effects at all if all of the proxies are already present, then we should move the proxies setup into the `use` command instead of install. There is a good possibility that there will be toolchain selectors for toolchains that aren't installed by swiftly coming in the future.

Sources/Swiftly/Unlink.swift
```diff
@@ -0,0 +1,57 @@
...
+        try await self.run(Swiftly.createDefaultContext())
+    }
+
+    mutating func run(_ ctx: SwiftlyCoreContext) async throws {
+        try await validatedConfig(ctx)
+
+        var pathChanged = false
+        if let proxyTo = try? await Swiftly.currentPlatform.findSwiftlyBin(ctx) {
+            let swiftlyBinDir = Swiftly.currentPlatform.swiftlyBinDir(ctx)
+            let swiftlyBinDirContents = (try? FileManager.default.contentsOfDirectory(atPath: swiftlyBinDir.string)) ?? [String]()
```
cmcgee1024, line 34: suggestion (non-blocking): You can use the fs (aka. FileSystem) `ls()` function for this and it accepts FilePath directly.

Sources/Swiftly/Unlink.swift
```diff
@@ -0,0 +1,57 @@
...
+
+            let existingProxies = swiftlyBinDirContents.filter { bin in
+                do {
+                    let linkTarget = try FileManager.default.destinationOfSymbolicLink(atPath: (swiftlyBinDir / bin).string)
```
cmcgee1024, line 38: suggestion (non-blocking): There's a `fs.readlink()` function that will read symlinks for you with a FilePath of the symlink.

Sources/Swiftly/Install.swift
cmcgee1024: On second thought, maybe proxies are set up in a variety of places: link, install, and use.

plemarquand: @cmcgee1024 After second thought, should this method move?

Sources/Swiftly/Unlink.swift
plemarquand: Good call, done!

Sources/Swiftly/Install.swift
cmcgee1024: Let's keep it here. It can always be moved later.

Sources/Swiftly/Unlink.swift
```diff
@@ -0,0 +1,58 @@
...
+        abstract: "Unlinks swiftly so it no longer manages the active toolchain."
+    )
+
+    @Argument(help: ArgumentHelp(
+        "Unlinks swiftly, allowing the system default toolchain to be used.",
+        discussion: """
+
+        Unlinks swiftly until swiftly is linked again with:
+
+            $ swiftly link
```
cmcgee1024, line 17: suggestion: Output this message at the end of the unlink command to tell users how they can link up to the active toolchain again.

thought: I wonder if other swiftly subcommands should indicate that the user is in an "unlinked" state and give them the message showing them how to get back. At least `swiftly use` is a command that people might use to see what's going on with swiftly, so that might be a good candidate. Detecting whether you're unlinked might be as simple as checking that there's an active toolchain and noticing that there are no proxies in the swiftly bin directory.

Sources/SwiftlyCore/Messages.swift
```diff
@@ -0,0 +1,22 @@
+public enum Messages {
```
cmcgee1024, line 1: praise: This is a good pattern for sharing common messages. The static function is a nice touch for interpolated messages.

Sources/SwiftlyCore/Messages.swift
```diff
@@ -0,0 +1,22 @@
...
+    NOTE: Swiftly has updated some elements in your path and your shell may not yet be
+    aware of the changes. You can update your shell's environment by running
+
+    hash -r
+
+    or restarting your shell.
+
+    """
+
+    public static func postInstall(_ message: String) -> String {
```
cmcgee1024, line 12: suggestion (non-blocking): I think the right name for this parameter is "command". It'll be a command that it asks the user to run themselves as an administrator.

Sources/SwiftlyCore/Platform.swift
```diff
@@ -454,11 +454,6 @@ extension Platform {
             return cmdAbsolute
         }
 
-        // If we're running inside an xctest then we don't have a location for this swiftly.
```
cmcgee1024, line 457: question: Why remove this block? Historically, it was really dangerous when running tests not to perform this check. It might even cause problems for a user's swiftly install after they run the test suite.

plemarquand: It leaves large portions of the link/unlink code untested (other commands too, I assume). If tests are using mocked IO and temporary directories is this still dangerous?

cmcgee1024: For a near-term suggestion, I recommend restoring this code and doing some brief link/unlink testing in the end-to-end tests. See here for where the E2E tests finally test that the `swift` proxy is working.

https://github.com/swiftlang/swiftly/blob/df0e91c3e851c4d779ac68de927fed6fecb5829f/Sources/TestSwiftly/TestSwiftly.swift#L96

Perhaps the test could add an unlink, and link, to check that `which swift` goes somewhere else, and then comes back afterwards at the end of everything?

Longer term, I do wonder if we could mock in a much safer way the proxies themselves. IIRC the proxies were being created at one point so that they symlinked to the test helper, or the user's swiftly from their PATH, which caused all sorts of problems. I'm not sure if you want to tackle that problem right now.

Sources/Swiftly/Unlink.swift
plemarquand: This is a good idea, I've updated `use`, `install` and `update` so they check if swiftly is unlinked. A warning message is printed preceding the command output if swiftly is not linked.

cmcgee1024 (commented):
Overall, this looks good. There's just the merge conflicts (another PR to provide gentle update messages).

cmcgee1024:
@swift-ci test macOS

AvdLee:
Any idea when this will be released? I'd love to have this feature!

PR #317, opened 2025-04-11
Create a special xcode selector that will delegates to xcrun for macOS

cmcgee1024:
Add the feature of a special selector "xcode" that can be used like the other swiftly
selectors for some operations, such as `swiftly use`, and `swiftly list`, but has no
real effect on other operations.

With this feature, you can `swiftly use xcode` and then when you run the proxies for
items like swift, clang, and other common toolchain binaries they will be run as the
equivalent `xcrun` command.

```
swift --version  => xcrun swift --version
clang foo.c => xcrun clang foo.c
```

When xcode is selected then `swiftly run` will run with the currently selected xcode
toolchain (or command-line tools) reported by `xcrun -f` as the toolchain on the path.

When you run `swiftly list` there is a special entry for "xcode" that is decorated in a
similar way to indicate whether it is the (global) default, and/or in-use toolchain as a
result of the selection mechanism.

TODO:
* Document the new selector in all of the usual places
* Condition the selector and everything that is related to it on the macOS platform so that Linux users do not see it
* Write specific test cases for this version, selector, and commands that take versions or selectors

### DIFF
--- a/Sources/MacOSPlatform/MacOS.swift
+++ b/Sources/MacOSPlatform/MacOS.swift
@@ -207,9 +207,15 @@ public struct MacOS: Platform {
         return "/bin/zsh"
     }
 
-    public func findToolchainLocation(_ ctx: SwiftlyCoreContext, _ toolchain: ToolchainVersion) -> URL
-    {
-        self.swiftlyToolchainsDir(ctx).appendingPathComponent("\(toolchain.identifier).xctoolchain")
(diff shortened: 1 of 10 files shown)


cmcgee1024:
@swift-ci test macOS

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
@swift-ci test macOS

PR #318, opened 2025-04-13
Improve path management and filesystem operation ergonomics

cmcgee1024:
Use FilePath instead of file URL's. FilePath is recommended as the system data type to be used to represent local file paths for command-line tools. The methods have simpler names while retaining the same vital path arithmetic functions. It is much less likely that a stringer will accidentally print out a file URL to the user when a path is intended.

Remove usage of 'FileManager.default' in favour of API's that are far less verbose to type and read. Make shorter and more command-line recognizable FileSystem API functions for operations, such as checking if a file exists, removing files, moving them, copying them, etc. These functions accept FilePath, not URL or String for a measure of type safety. Make the new API's async by default to permit swapping FileManager with another implementation that has async operations.

The most common file path operation is appending. Make use of operator overloading to make these operations much cleaner, and clearer with the division operator.


cmcgee1024:
@swift-ci test macOS

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
@swift-ci test macOS

Sources/SwiftlyCore/FileManager+FilePath.swift
```diff
@@ -0,0 +1,179 @@
...
+        try self.write(to: URL(fileURLWithPath: path.string), atomically: atomically, encoding: enc)
+    }
+
+    public init(contentsOf path: FilePath, encoding enc: String.Encoding = .utf8) throws {
+        try self.init(contentsOf: URL(fileURLWithPath: path.string), encoding: enc)
+    }
+}
+
+extension FilePath {
+    public static func / (left: FilePath, right: String) -> FilePath {
```
plemarquand, line 176: I don't love this kind of pseudo operator overloading. To me it obscures the intent vs plainly calling `left.append(right)`. 

In general Swift maintains that an overloaded operator should be consistent with the intent, so overloading `+` to work with a `Point` type makes sense since adding together the Point's coordinates is a natural operation consistent with the intent of addition. This meant that on first read I interpreted these usages as "divide the one segment in to the path into the other" before double taking and realizing that doesn't make sense.

Sources/SwiftlyCore/FileManager+FilePath.swift
```diff
@@ -0,0 +1,179 @@
...
+
+public func readlink(atPath: FilePath) async throws -> FilePath {
+    try FileManager.default.destinationOfSymbolicLink(atPath: atPath)
+}
+
+public func symlink(atPath: FilePath, linkPath: FilePath) async throws {
+    try FileManager.default.createSymbolicLink(atPath: atPath, withDestinationPath: linkPath)
+}
+
+public func chmod(atPath: FilePath, mode: Int) async throws {
```
plemarquand, line 91: Is there a reason these are defined as free functions instead of extensions on FileManager? Seems like some vars like `homeDir` are implemented in two places.

Sources/SwiftlyCore/Platform.swift
```diff
@@ -56,31 +57,31 @@ public struct RunProgramError: Swift.Error {
 public protocol Platform: Sendable {
     /// The platform-specific default location on disk for swiftly's home
     /// directory.
-    var defaultSwiftlyHomeDirectory: URL { get }
+    var defaultSwiftlyHomeDir: FilePath { get }
```
plemarquand, line 60: praise: nice consistency improvement

Sources/SwiftlyCore/FileManager+FilePath.swift
cmcgee1024: There's a tradeoff here in this case. I think that like with Python's pathlib package, people are rarely expecting that there's some kind of mathematical divide happening between file paths as they might if the data structure was more mathematical in nature. With path operations, there's a clear benefit to being able to see the paths, more than the method names, and Swift types involved so that one can reason about them at a higher level.

Sources/SwiftlyCore/FileManager+FilePath.swift
plemarquand: Generally I persist a local var like `let fm = FileManager.default` to make these calls more concise (i.e. `fm.chmod(...)`). This also lets you more easily swap out the FileManager implementation in tests to, say, an in memory file system. See something like [FileSystem in tools-support-core](https://github.com/swiftlang/swift-tools-support-core/blob/main/Sources/TSCBasic/FileSystem.swift#L177).

Sources/SwiftlyCore/FileManager+FilePath.swift
plemarquand: Fair, its a valid trade off, I've been bitten by overloaded operators before but in this case its probably pretty benign.

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
@swift-ci test macOS

plemarquand (approved):

PR #321, opened 2025-04-14
Remove swiftly from the path during proxying

cmcgee1024:
Retaining swiftly in the path increases the probability that a tool might run the proxy again, resulting in circularities. Also, tools will locate programs in the path to try and find adjacent toolchain files that cannot be discovered in the swiftly bin directory.

Remove the swiftly bin directory from the path when proxying to help improve the experience with other tools.


cmcgee1024:
@swift-ci test macOS

plemarquand (approved):

Issue #323, opened 2025-04-16
Swiftly cannot be installed in a headless macOS


cmcgee1024:
When installing the swiftly pkg on a headless macOS that has no GUI at all, there is the following error from the installer:

```
% installer -pkg swiftly.pkg -target CurrentUserHomeDirectory -verbose
installer: Package name is
installer: Installing at base path /Users/github
installer: Preparing for installation….....
installer: Preparing the disk….....
installer: Preparing ….....
installer: The install failed..
```

A similar problem happens when swiftly tries to install a swift toolchain into the user home directory.

cmcgee1024:
It is possible to extract swiftly on a headless macOS if swiftly is checked, and extracted using pkgutil like this, and the toolchain directory is changed from the default so that swiftly itself uses the same mechanism:

```
export SWIFTLY_HOME_DIR=~/.swiftly
export SWIFTLY_BIN_DIR="${SWIFTLY_HOME_DIR}/bin"
export SWIFTLY_TOOLCHAINS_DIR="${SWIFTLY_HOME_DIR}/toolchains"
export SWIFTLY_TEMP=$(mktemp -d)
pkgutil --check-signature swiftly.pkg
pkgutil --expand swiftly.pkg "$SWIFTLY_HOME_DIR"
tar -C "${SWIFTLY_HOME_DIR}" -xf "${SWIFTLY_TEMP}/swiftly-*.pkg/Payload"
"${SWIFTLY_BIN_DIR}/swiftly" init
```

PR #324, opened 2025-04-16
Make the toolchains directory location configurable

cmcgee1024:
Using a new environment variable SWIFTLY_TOOLCHAINS_DIR that follows a similar pattern as SWIFTLY_HOME_DIR and SWIFTLY_BIN_DIR make it possible for the user to configure the toolchain location on init in the same way.

On macOS, this level of configuration requires a different approach to extracting the toolchain whenever the installation location is anything other than the default. Use the pkgutil utility of macOS to extract the toolchain in the location specified by the user. Installing the toolchains outside of the installer in a custom location means that Xcode may not be able to pick them up easily. Make a note of that in the init screen so that users are aware.

When using the init `--no-modify-profile` flag the init was skiping steps like installing the latest toolchain, and emitting the notes about updating the current shell environment. Separate this logic so that the user can have these steps performed with the flag.

This can help with #323 if you use `pkgutil --extract` and `pkgutil --check-signature` to extract swiftly, and then swiftly will install toolchains using a similar mechanism, provided that you customize the `SWIFTLY_TOOLCHAINS_DIR` to something other than the default.

### DIFF
--- a/Sources/LinuxPlatform/Linux.swift
+++ b/Sources/LinuxPlatform/Linux.swift
@@ -28,7 +28,7 @@ public struct Linux: Platform {
     }
 
     public func swiftlyBinDir(_ ctx: SwiftlyCoreContext) -> URL {
-        ctx.mockedHomeDir.map { $0.appendingPathComponent(".local/share/swiftly/bin", isDirectory: true) }
+        ctx.mockedHomeDir.map { $0.appendingPathComponent("bin", isDirectory: true) }
             ?? ProcessInfo.processInfo.environment["SWIFTLY_BIN_DIR"].map { URL(fileURLWithPath: $0) }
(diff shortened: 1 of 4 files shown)


cmcgee1024:
@swift-ci test macOS

justice-adams-apple (approved):
LGTM do we want to add a test which sets the var?

cmcgee1024:
> LGTM do we want to add a test which sets the var?

In a sense they do test the var because the tests exc

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
> LGTM do we want to add a test which sets the var?

I've added a new E2E test that installs swiftly in a custom location with a custom toolchains directory.

PR #326, opened 2025-04-18
Make the toolchains directory location configurable (#324)

cmcgee1024:
Using a new environment variable SWIFTLY_TOOLCHAINS_DIR that follows a similar pattern as SWIFTLY_HOME_DIR and SWIFTLY_BIN_DIR make it possible for the user to configure the toolchain location on init in the same way.

On macOS, this level of configuration requires a different approach to extracting the toolchain whenever the installation location is anything other than the default. Use the pkgutil utility of macOS to extract the toolchain in the location specified by the user. Installing the toolchains outside of the installer in a custom location means that Xcode may not be able to pick them up easily. Make a note of that in the init screen so that users are aware.

When using the init `--no-modify-profile` flag the init was skiping steps like installing the latest toolchain, and emitting the notes about updating the current shell environment. Separate this logic so that the user can have these steps performed with the flag.

### DIFF
--- a/Sources/LinuxPlatform/Linux.swift
+++ b/Sources/LinuxPlatform/Linux.swift
@@ -37,7 +37,9 @@ public struct Linux: Platform {
     }
 
     public var swiftlyToolchainsDir: URL {
-        self.swiftlyHomeDir.appendingPathComponent("toolchains", isDirectory: true)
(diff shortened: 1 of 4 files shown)


cmcgee1024:
@swift-ci test macOS

PR #340, opened 2025-04-30
Add a guide for accessing tools installed by swiftly and other places

cmcgee1024:
Document how one can access tools outside of the usual swiftly
proxy mechanism for various purposes. One possibility is to
determine if there is a bug in a particular tool. Another is that
there is a tool that is installed outside of swiftly.

### DIFF
--- a/Documentation/SwiftlyDocs.docc/use-tools-different-sources.md
+++ b/Documentation/SwiftlyDocs.docc/use-tools-different-sources.md
@@ -1,0 +1,39 @@
+# Access Tools from Different Sources
+
+Access tools that can be from different sources.
+
+swiftly installs and manages toolchains for you that are available from swift.org. It sets up your shell environment to work well with these. There are other sources of Swift toolchains, and the common tools found in them, too. How do I access these from within my swiftly environment?
+
+Also, sometimes it's difficult to determine whether there is a bug in swiftly itself, or the bug lies in the particular toolchain at swift.org. How can I figure this out?
+
+If the tool is in a toolchain managed by swiftly it will call it directly when you run "swift", "clang", or "lldb". If you want to access the particular tool directly you can use `swiftly run` to run it like this wit the particular version of toolchain that you want:
+
+```
+swiftly run clang --version
+```
+
+If you want to bypass swiftly's run mechanism you can find the location of the tool and run it from it's fully qualified path:
+
+```
+$(swiftly use --print-location)/usr/bin/clang --version
+```
+
+Normally, when you install swiftly it places a set of proxy tools for the usual Swift toolchain commands. These are run in place of the actual toolchain so that swiftly can route them to the in use toolchain based on your configuration. If you want to bypass swiftly entirely, or use tools that were installed separately from swiftly there is also the ability to "unlink" swiftly's proxy binaries from your PATH.
+
+```
+swiftly unlink
+```
+
+Now, when you run "swift" and other tools they will be found elsewhere on your system (or not found at all). The "swiftly" command should still be available, so you can still find the location of the in use toolchain and other swiftly commands.
+
+```
+swiftly use --print-location
+```
+
+For instance, you can use this path to call a tool directly, or manually construct your path with it. You can also use swiftly to link the proxies again to get back to where you were before:
+
+```
+swiftly link
+```
+
+In this guide you have seen a few different ways to access tools installed by swiftly, and ones that are installed separately.


Documentation/SwiftlyDocs.docc/use-tools-different-sources.md
```diff
@@ -0,0 +1,39 @@
+# Access Tools from Different Sources
+
+Access tools that can be from different sources.
+
```
heckj, line 4:
```suggestion

## Overview
```

Documentation/SwiftlyDocs.docc/use-tools-different-sources.md
```diff
@@ -0,0 +1,39 @@
...
+swiftly installs and manages toolchains for you that are available from swift.org. It sets up your shell environment to work well with these. There are other sources of Swift toolchains, and the common tools found in them, too. How do I access these from within my swiftly environment?
+
+Also, sometimes it's difficult to determine whether there is a bug in swiftly itself, or the bug lies in the particular toolchain at swift.org. How can I figure this out?
```
heckj, line 7: This articles seems to be trying to serve two masters - how to use the new `link` and `unlink` subcommands as well as how to wrangle seeing and controlling where a specific tool is being invoked from. I think you may be trying to go too broad here. This reads to me like two focused articles sort of mashed into one.

Documentation/SwiftlyDocs.docc/use-tools-different-sources.md
```diff
@@ -0,0 +1,39 @@
...
+```
+swiftly run clang --version
+```
+
+If you want to bypass swiftly's run mechanism you can find the location of the tool and run it from it's fully qualified path:
+
+```
+$(swiftly use --print-location)/usr/bin/clang --version
+```
+
```
heckj, line 20:
```suggestion
### Link and Unlink Swiftly's command proxy
```

Documentation/SwiftlyDocs.docc/use-tools-different-sources.md
```diff
@@ -0,0 +1,39 @@
+# Access Tools from Different Sources
+
+Access tools that can be from different sources.
+
+swiftly installs and manages toolchains for you that are available from swift.org. It sets up your shell environment to work well with these. There are other sources of Swift toolchains, and the common tools found in them, too. How do I access these from within my swiftly environment?
+
+Also, sometimes it's difficult to determine whether there is a bug in swiftly itself, or the bug lies in the particular toolchain at swift.org. How can I figure this out?
+
```
heckj, line 8:
```suggestion

### Use the run subcommand to invoke executables in a Swift toolchain
```

This sort of overlaps with content you have in `use-toolchains.md` - maybe it's worth breaking this out into it's own article on using `swiftly run` - since you can do quite a bit with it - and referencing that article from here?

Documentation/SwiftlyDocs.docc/use-tools-different-sources.md
```diff
@@ -0,0 +1,39 @@
...
+If the tool is in a toolchain managed by swiftly it will call it directly when you run "swift", "clang", or "lldb". If you want to access the particular tool directly you can use `swiftly run` to run it like this wit the particular version of toolchain that you want:
```
heckj, line 9:
```suggestion
If the tool is in a toolchain managed by swiftly it calls it directly when you run "swift", "clang", or "lldb".
If you want to access the particular tool directly, use `swiftly run` to run it with the toolchain  you choose.
The following example illustrates how to invoke `clang --version` from the currently enabled toolchain:
```

Documentation/SwiftlyDocs.docc/use-tools-different-sources.md
```diff
@@ -0,0 +1,39 @@
...
+
+```
+swiftly run clang --version
+```
+
+If you want to bypass swiftly's run mechanism you can find the location of the tool and run it from it's fully qualified path:
```
heckj, line 15:
```suggestion
If you want to bypass swiftly's run mechanism, use `swiftly use --print-location` to find the location of the tool and run it from there.
The following example shows how to use this within a shell to create a full path to the currently active toolchain:
```

Documentation/SwiftlyDocs.docc/use-tools-different-sources.md
```diff
@@ -0,0 +1,39 @@
...
+
+```
+$(swiftly use --print-location)/usr/bin/clang --version
+```
+
+Normally, when you install swiftly it places a set of proxy tools for the usual Swift toolchain commands. These are run in place of the actual toolchain so that swiftly can route them to the in use toolchain based on your configuration. If you want to bypass swiftly entirely, or use tools that were installed separately from swiftly there is also the ability to "unlink" swiftly's proxy binaries from your PATH.
```
heckj, line 21:
```suggestion
Normally, swiftly installs a set of proxy commands on your `PATH` environment variable for common Swift toolchain executables when you select a toolchain for use.
These are run in place of the actual toolchain so that swiftly can route them to the in use toolchain based on your configuration.
If you want to bypass swiftly entirely, or use tools that were installed separately from swiftly there is also the ability to "unlink" the proxy binaries.
```

Documentation/SwiftlyDocs.docc/use-tools-different-sources.md
```diff
@@ -0,0 +1,39 @@
...
+
+```
+swiftly unlink
+```
+
+Now, when you run "swift" and other tools they will be found elsewhere on your system (or not found at all). The "swiftly" command should still be available, so you can still find the location of the in use toolchain and other swiftly commands.
```
heckj, line 27:
```suggestion
After you run `swiftly unlink`,  the system will search your PATH for `swift` and other executables in the toolchain without the proxy commands included.
The `swiftly` command is still be available, which you can use to find and run executables within any toolchain that swiftly installed.
```

if `swiftly run` was its own article, I'd do a "For more information on using `swiftly` to invoke executables from toolchains, read `<doc:swiftly-run-article>`.

Documentation/SwiftlyDocs.docc/use-tools-different-sources.md
```diff
@@ -0,0 +1,39 @@
...
+
+```
```
heckj, line 29: lines 29-31 seem to be repeat of what you're describing up at line 18 - I think you'd want to either move that content down here and flesh this out, or push this detail back up outside of the section on link and unlink.

Documentation/SwiftlyDocs.docc/use-tools-different-sources.md
```diff
@@ -0,0 +1,39 @@
...
+swiftly use --print-location
+```
+
+For instance, you can use this path to call a tool directly, or manually construct your path with it. You can also use swiftly to link the proxies again to get back to where you were before:
```
heckj, line 33:
```suggestion
After you unlinked the command proxies, run the following command to enable them again:
```

heckj (changes requested):
I'll repeat this below in the commands, but I think you've got two articles here mashed into one. I suggested added section headers to try and break up the content into more concrete flows, but even with these suggestions the article seems to be serving two masters. 

I made suggestions through the content to tighten up wording or rephrase to make the wording a bit more targeted. you're describing two different procedural things, and tacking in a bit of explanation in between. At a high level, what I'm seeing:

- how to use (the new) link and unlink subcommands
- a bit of explanation that swiftly adds executable proxies for common commands to your PATH
- how to use `swiftly run` and `swiftly use --print-location` to either invoke within a toolchain, or assemble the full path to do so yourself.

PR #344, opened 2025-05-05
Add ability to self uninstall swiftly

### LINKED ISSUE #84: Add `self-uninstall` command
We should make it easy to cleanly uninstall swiftly in case a user no longer needs it on their system.

louisunlimited:
## `swiftly self-uninstall`
- Fixes #84 

Implements the command `swiftly self-uninstall`, the ability to self uninstall swiftly incase it's no longer needed.
It deletes swiftly by removing everything under `swiftBinDir` and `swiftHomeDir`, and also removes the lines in shell profiles added by swiftly

## TODO:
- [x] Implement removing shell profile entry
- [x] Tests for removing shell profile entry
- [x] Update documentation in README
- [x] Update docc ref

## Questions:
- Are swift bin and swift home dir guaranteed to be safe for removal with `fs.remove(atPath: ...)`?
- Anything else we need to delete?
- Specific requirements for the deletion flow?
- What flags should we support?

## Issue
Due to the way we mock home directories, the `removesEntryFromShellProfile` fails because any shell profiles `.zprofile`, `.profile` will be removed alongside `swiftlyBinDir`/`swiftlyHomeDir`, so the expect present check fails. Need to come up a better way to manage `mockedTestHome` directories


Sources/Swiftly/SelfUninstall.swift
```diff
@@ -0,0 +1,119 @@
...
+        } else {
+            envFile = Swiftly.currentPlatform.swiftlyHomeDir(ctx) / "env.sh"
+            sourceLine = """
+
+            # Added by swiftly
+            . "\(envFile)"
+            """
+        }
+
+        let userHome = ctx.mockedHomeDir ?? fs.home
```
cmcgee1024, line 76: suggestion (blocking): The current state of the user's shell may not be the same as when they initially installed swiftly. In the future, swiftly might actually install itself into all supported shells because users (like myself) switch shells quite often. So, instead of picking the current profileHome, this code should check all possible profiles for the source lines, and remove them.

Tests/SwiftlyTests/SelfUninstallTests.swift
```diff
@@ -0,0 +1,129 @@
+import Foundation
+@testable import Swiftly
+@testable import SwiftlyCore
+import SystemPackage
+import Testing
+
+@Suite struct SelfUninstallTests {
```
cmcgee1024, line 7: praise: It's good to see the tests here for this functionality that are small and fast to run.

Tests/SwiftlyTests/SelfUninstallTests.swift
```diff
@@ -0,0 +1,129 @@
...
+        "/bin/fish",
+    ]) func removesEntryFromShellProfile(_ shell: String) async throws {
+        var ctx = SwiftlyTests.ctx
+        ctx.mockedShell = shell
+
+        try await SwiftlyTests.$ctx.withValue(ctx) {
+            // Create a profile file with the source line
+            let userHome = SwiftlyTests.ctx.mockedHomeDir!
+
+            let profileHome: FilePath
```
cmcgee1024, line 47: question: Can this test invoke a `swiftly init` to do the usual install behaviour instead of duplicating the logic here?

cmcgee1024 (changes requested):
Overall, this looks good. I've added a few suggestions, and a question.

Sources/Swiftly/SelfUninstall.swift
```diff
@@ -0,0 +1,119 @@
...
+
+    public static func execute(_ ctx: SwiftlyCoreContext, verbose _: Bool) async throws {
+        await ctx.print("""
+        You are about to uninstall swiftly. 
+        This will remove the swiftly binary and all the files in the swiftly home directory. 
+        All installed toolchains will not be removed, if you want to remove them, please do so manually with `swiftly uninstall all`.
+        This action is irreversible.
+        """)
+
+        guard await ctx.promptForConfirmation(defaultBehavior: true) else {
```
cmcgee1024, line 43: issue: Since this is a prompt for confirmation, it should also support an "--assume-yes" override in case of an automated workflow that might install and then uninstall swiftly afterwards.

louisunlimited:
Hi @cmcgee1024 I just made the changes you requested! lmk if there's anything else :)

README.md
```diff
@@ -59,7 +59,12 @@ This command checks to see if there are new versions of swiftly itself and upgra
 
 ## Uninstalling swiftly
 
-Currently, only manual uninstallation is supported. If you need to uninstall swiftly, please follow the instructions below:
+swiftly can be savely removed with the following command:
```
cmcgee1024, line 62: nitpick: `s/savely/safely/` here

Sources/Swiftly/SelfUninstall.swift
```diff
@@ -0,0 +1,103 @@
...
+            let sourceLine = isFish ? fishSourceLine : shSourceLine
+            let contents = try String(contentsOf: path, encoding: .utf8)
+            if contents.contains(sourceLine) {
+                let updated = contents.replacingOccurrences(of: sourceLine, with: "")
+                try Data(updated.utf8).write(to: path, options: [.atomic])
+            }
+        }
+
+        await ctx.print("Removing swiftly binary at \(swiftlyBin)...")
+        try await fs.remove(atPath: swiftlyBin)
```
cmcgee1024, line 96: issue (blocking): Thinking on this a little bit I think that we will need to take some extra care here. `fs.remove` is a big hammer. It can remove important directories and files without warning or recourse. The user can set their `SWIFTLY_BIN_DIR` to a shared directory with other binaries in it.

suggestion: Instead of removing the directory, how about removing the individual proxies by checking if they are symlinks to the swiftly binary, then remove the swiftly binary itself. Finally, if this directory is empty, then remove it.

The same kind of thing should probably be used for the `SWIFTLY_HOME_DIR` too. Remove the config.json, the environment files, and then remove the directory if it is empty.

For toolchains directory, remove all of the known toolchains first, and then check if it is empty.

Tests/SwiftlyTests/SelfUninstallTests.swift
```diff
@@ -0,0 +1,96 @@
...
+    }
+
+    func shellProfileRemovalTest() async throws {
+        try await SwiftlyTests.withTestHome {
+            // Fresh user without swiftly installed
+            try? await fs.remove(atPath: Swiftly.currentPlatform.swiftlyConfigFile(SwiftlyTests.ctx))
+            try await SwiftlyTests.runCommand(Init.self, ["init", "--assume-yes", "--skip-install"])
+
+            let fishSourceLine = """
+            # Added by swiftly
```
cmcgee1024, line 54: thought: The user might reorganize their profile manually, or invoke some kind of tooling that does that. Some of these lines might exist, and others might not. Maybe a user removes the comment, leaving only the source line. We could perhaps match each line and remove them individually to make this a bit more robust and safe. Each one is unique enough that I think they shouldn't have too many false matches.

Sources/Swiftly/SelfUninstall.swift
cmcgee1024: Let's take advantage of the verbose flag here. If it's set then we report every file path that we removed.

louisunlimited: Are we making this self-uninstall in a way that it also deletes the toolchains directory? I was thinking it might be better to delegate this to the user since they might still want to use the toolchains, just not with swiftly managing them.

louisunlimited:
Hi @cmcgee1024 Just updated the code, lmk!

louisunlimited:
Hi @cmcgee1024 could you take a look at this when you have time? Thanks!

Sources/Swiftly/SelfUninstall.swift
cmcgee1024: I think that you're right. Toolchains are large enough to download and place on disk. The user might want to manage these separately. The note does seem to clearly state to the user what's going to happen, giving them a chance to abort and uninstall all of the toolchains.

README.md
louisunlimited: This should already been fixed by commit 823de2fa7b993c3386aaddafa50c8a6bd0fca331

cmcgee1024:
@louisunlimited there's a test failure on Linux here: https://github.com/swiftlang/swiftly/actions/runs/18237331411/job/51948714037?pr=344#step:5:3471

Also, you can fix the formatting with `swift run swiftformat .`

louisunlimited:
I'm guessing it's because the `Toolchains` path is called `toolchains` on linux, can you run the test again and see?

PR #345, opened 2025-05-07
Generate command models from argument parser info JSON files

cmcgee1024:
Instead of hand-crafting the helper functions, options, and structs for each command,
describe the command structure using an argument parser compatible JSON format
for each command. Introduce a generator plugin that generates all of the Swift types in
a manner that is similar to the way that they were hand-written before.

There are certain limitations in the JSON schema for the Swift argument parser that
prevent the generator from knowing certain important details. For example, there is
no information on which arguments contain file paths, which is important information
for type safety in this context. For this information, and more potentially in the future
introduce a "-ext.json" file with an extension schema to hold the extra information.


cmcgee1024:
@swift-ci test macOS

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
@swift-ci test macOS

cmcgee1024:
@swift-ci test macOS

PR #350, opened 2025-05-12
Add support for sudo installation missing packages after init

cmcgee1024:
Currently when there is a post installation command during init the user is given the command that they must run at the end. In order to streamline the process further an additional flag called '--sudo-install-packages' is added that will invoke sudo on behalf of the user to perform that command as root.

Add the new flag to the init subcommand. Create a regex that restricts the allowable commands to a narrow set of patterns as a measure of protection. Invoke the sudo process directly from the expected directory on supported Linux systems, which is /usr/bin/sudo


cmcgee1024:
@swift-ci test macOS

plemarquand (approved):

PR #352, opened 2025-05-13
Add --install-system-deps flag to automatically install system dependencies

justice-adams-apple:
Add an install option `--install-system-deps` or `-i` so that users can auto-install the system dependencies required to install a toolchain on the given platform. This should help the install experience for docker users


justice-adams-apple:
cc @cmcgee1024 @shahmishal

PR #355, opened 2025-05-14
Append file extension to downloaded toolchain

plemarquand:
The `installer` tool fails to install the toolchain on macOS unless `.pkg` is present in the filename. Otherwise it produces the error: "Error - the package path specified was invalid"

Add the appropriate file extension when downloading the toolchain so the installer can use it correctly.


cmcgee1024:
@swift-ci test macOS

PR #356, opened 2025-05-14
Cleanup Uninstall message

plemarquand:
The uninstall message would print:
`Uninstalling main-snapshot-2025-05-12...Uninstalling package in user home directory...`

Clean it up so the second half only prints with `--verbose`. Also clean up a test assertion to make it more specific.


cmcgee1024:
@swift-ci test macOS

Issue #142, opened 2024-07-21
Add lock to ensure only one Swiftly process is running at any point


adam-fowler:
Running two swiftly processes at the same produces undefined results. 
We should add a file lock to ensure only one swiftly process is touching the swiftly folder structure at any one point

bkhouri:
I encountered this also while running `swiftly update` simultaneously in two different terminal sessions.  The command was installing the same toolchain version in both instances.

roulpriya:
@adam-fowler Can I work on this issue?

adam-fowler:
I'm not really that involved in the swiftly project anymore. @cmcgee1024 can make a call on that.

cmcgee1024:
@roulpriya I don't think that anyone else is working on this at the moment. I'm assigning this to you to indicate that you're taking a look at it right now.

roulpriya:
@cmcgee1024 I came across how SwiftPM implements locking.

When we try to execute multiple instances of SwiftPM we get a log like 
`Another instance of SwiftPM is already running using '/Users/priyaroul/Code/swiftly/.build', waiting ...`

In Swiftly, should we wait to acquire the lock or print an error and exit?

cmcgee1024:
In my opinion lock acquisition should exit immediately with a specific exit code that scripts (or VSCode) can check and decide on their polling, and/or timeout strategy. Those things will have a better understanding of their environment, and the constraints.

The proxies, and `swift run` should not be interrupted by a lock since they have no good way to communicate the lock acquisition failure, limited by the behaviour of the commands that they proxy. They can be optimistic about presence of the config.json, and the selected toolchain. Uninstallation of a toolchain is also likely a relatively rare operation.

cmcgee1024:
This work is complete with the linked PR's.

PR #361, opened 2025-05-18
Adding FileLock

roulpriya:
Fixes Issue #142 

- **Description**:
Serialize concurrent swiftly install commands with a simple file lock.
Adds a lightweight FileLock (using `Data.write(to: fileURL, options: .withoutOverwriting)`) and wraps the install routine so only one process installs at a time

### DIFF
--- a/Sources/Swiftly/Install.swift
+++ b/Sources/Swiftly/Install.swift
@@ -357,7 +357,7 @@ struct Install: SwiftlyCommand {
 
             let (pathChanged, newConfig) = try await withLock(lockFile) {
                 if verbose {
-                    await ctx.print("Acquired installation lock.")
+                    await ctx.print("Acquired installation lock")
                 }
 
                 var config = try await Config.load(ctx)
--- a/Sources/Swiftly/Install.swift
+++ b/Sources/Swiftly/Install.swift
@@ -81,12 +81,42 @@ struct Install: SwiftlyCommand {
         try await self.run(Swiftly.createDefaultContext())
     }
 
+    private func swiftlyHomeDir(_ ctx: SwiftlyCoreContext) -> FilePath {
+        Swiftly.currentPlatform.swiftlyHomeDir(ctx)
+    }
+
     mutating func run(_ ctx: SwiftlyCoreContext) async throws {
         let versionUpdateReminder = try await validateSwiftly(ctx)
         defer {
             versionUpdateReminder()
         }
         try await validateLinked(ctx)
+
+        let lockFile = self.swiftlyHomeDir(ctx) / "swiftly.lock"
+        let fileLock: SwiftlyCore.FileLock
+
+        do {
+            fileLock = try SwiftlyCore.FileLock(at: lockFile.string)
+        } catch {
+            print("ERROR: Failed to create lock file: \(error)")
+            Foundation.exit(42)
+        }
+
+        defer {
+            do {
+                try fileLock.unlock()
+            } catch {
+                print("WARNING: Failed to unlock file: \(error)")
+            }
+        }
+
+        if self.root.verbose {
+            await ctx.print("Attempting to acquire installation lock...")
+        }
+        guard fileLock.waitForLock(timeout: 300, pollingInterval: 2) else {
+            print("ERROR: Failed to acquire lock on file: \(lockFile.string) after 300 seconds")
+            Foundation.exit(42)
+        }
 
         var config = try await Config.load(ctx)
         let toolchainVersion = try await Self.determineToolchainVersion(ctx, version: self.version, config: &config)
--- a/Sources/Swiftly/Install.swift
+++ b/Sources/Swiftly/Install.swift
@@ -92,32 +92,6 @@ struct Install: SwiftlyCommand {
         }
         try await validateLinked(ctx)
 
-        let lockFile = self.swiftlyHomeDir(ctx) / "swiftly.lock"
-        let fileLock: SwiftlyCore.FileLock
-
-        do {
-            fileLock = try SwiftlyCore.FileLock(at: lockFile.string)
-        } catch {
-            print("ERROR: Failed to create lock file: \(error)")
-            Foundation.exit(42)
-        }
-
-        defer {
-            do {
-                try fileLock.unlock()
-            } catch {
-                print("WARNING: Failed to unlock file: \(error)")
-            }
-        }
-
-        if self.root.verbose {
-            await ctx.print("Attempting to acquire installation lock...")
-        }
-        guard fileLock.waitForLock(timeout: 300, pollingInterval: 2) else {
-            print("ERROR: Failed to acquire lock on file: \(lockFile.string) after 300 seconds")
-            Foundation.exit(42)
-        }
-
         var config = try await Config.load(ctx)
         let toolchainVersion = try await Self.determineToolchainVersion(ctx, version: self.version, config: &config)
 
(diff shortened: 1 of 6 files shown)


Sources/Swiftly/Install.swift
```diff
@@ -81,13 +81,43 @@ struct Install: SwiftlyCommand {
...
         defer {
             versionUpdateReminder()
         }
         try await validateLinked(ctx)
 
+        let lockFile = self.swiftlyHomeDir(ctx) / "swiftly.lock"
+        let fileLock: SwiftlyCore.FileLock
+
+        do {
+            fileLock = try SwiftlyCore.FileLock(at: lockFile.string)
```
cmcgee1024, line 99: question: Can the `at` parameter here take a FilePath instead of a String? Then anyone who uses it can get the benefit of the FilePath API, such as the path arithmetic operations.

Sources/SwiftlyCore/FileLock.swift
```diff
@@ -0,0 +1,99 @@
...
+    let fileHandle: FileHandle
+
+    public static let defaultPollingInterval: TimeInterval = 1.0
+
+    public static let defaultTimeout: TimeInterval = 300.0
+
+    public init(at filePath: String) throws {
+        self.filePath = filePath
+
+        if !FileManager.default.fileExists(atPath: filePath) {
```
cmcgee1024, line 27: suggestion: SwiftlyCore has some API's for doing simple file operations in a way that uses FilePath, and also make it independent of the FileManager. There are GH issues about adopting something like the NIOFileSystem, which are asynchronous. That API might help and make it easier to switch to it since the API's are all asynchronous too.

Sources/SwiftlyCore/FileLock.swift
```diff
@@ -0,0 +1,99 @@
...
+    let filePath: String
+
+    let fileHandle: FileHandle
+
+    public static let defaultPollingInterval: TimeInterval = 1.0
+
+    public static let defaultTimeout: TimeInterval = 300.0
+
+    public init(at filePath: String) throws {
+        self.filePath = filePath
```
cmcgee1024, line 25: question: This lock appears to have two stages to it. First is to create the file (if it doesn't exist), and then use flock() to create a file system lock. Can this be reduced to only one stage, which is creating the file?

roulpriya: Doing so has one issue, if the locking process is dead, the lock will not be released automatically. However with `flock()` we can guarantee that the lock will be released.

Sources/SwiftlyCore/FileLock.swift
```diff
@@ -0,0 +1,80 @@
...
+    timeout: TimeInterval = FileLock.defaultTimeout,
+    pollingInterval: TimeInterval = FileLock.defaultPollingInterval,
+    action: @escaping () async throws -> T
+) async throws -> T {
+    let lock = FileLock(at: lockFile)
+    guard try await lock.waitForLock(timeout: timeout, pollingInterval: pollingInterval) else {
+        throw SwiftlyError(message: "Failed to acquire file lock at \(lock.filePath)")
+    }
+
+    defer {
```
cmcgee1024, line 69: issue: It is super unfortunate that Swift doesn't yet have an async defer capability from an async function, which Go has. This means that this function might return and the lock file still exists for a time, which isn't what I would expect from a withXYZ style of function.

suggestion: Try this instead:

```
let retVal = try await action()
try await lock.unlock()
return retVal
```

Sources/SwiftlyCore/FileLock.swift
```diff
@@ -0,0 +1,80 @@
...
+
+/**
+ * A non-blocking file lock implementation using file creation as locking mechanism.
+ * Use case: When installing multiple Swiftly instances on the same machine,
+ * one should acquire the lock while others poll until it becomes available.
+ */
+
+public actor FileLock {
+    let filePath: FilePath
+    private var isLocked = false
```
cmcgee1024, line 12: suggestion: With Swift there's a preference for value types, such as structs, over reference types (actor, class). The idea is to avoid the overhead of reference counting, and faster concurrent access. How about we model this instead using a struct that can only be constructed once you've got a lock?

```
public struct FileLock {
    let filePath: FilePath

    public init() async throws {
        // This basically does what waitForLock() and constructs a FileLock only if the lock is acquired, otherwise it throws an error
    }
    
    public unlock() async throws {
        // Don't worry about the isLocked, just go ahead and try to remove the file.
        // Errors shouldn't happen here normally since we already have the "lock", so the file should still be there, but if an error is produced then we should report that
        try await FileSystem.remove(atPath: self.filePath)
    }
}
```

Sources/Swiftly/Install.swift
```diff
@@ -346,33 +350,51 @@ struct Install: SwiftlyCommand {
...
 
-            let pathChanged = try await Self.setupProxies(
-                ctx,
-                version: version,
-                verbose: verbose,
-                assumeYes: assumeYes
-            )
+            let (pathChanged, newConfig) = try await withLock(lockFile) {
+                if verbose {
+                    await ctx.print("Acquired installation lock.")
```
cmcgee1024, line 360: suggestion (non-blocking): If we're being verbose, then the user might want to see where the lock file is in case they're having to manually delete it or something. Can you add the path in this message?

cmcgee1024:
@swift-ci test macOS

Sources/Swiftly/Install.swift
roulpriya: I have already added it to the earlier log line 355. Should we add it twice?

Sources/SwiftlyCore/FileLock.swift
```diff
@@ -0,0 +1,74 @@
...
+    let filePath: FilePath
+
+    public static let defaultPollingInterval: TimeInterval = 1
+    public static let defaultTimeout: TimeInterval = 300.0
+
+    public init(at path: FilePath) throws {
+        self.filePath = path
+        do {
+            let fileURL = URL(fileURLWithPath: self.filePath.string)
+            let contents = getpid().description.data(using: .utf8) ?? Data()
```
jakepetroules, line 22: Can you use https://developer.apple.com/documentation/foundation/processinfo/processidentifier instead of `getpid()`? One fewer obstacle towards getting swiftly to work on Windows.

roulpriya: Refactored @jakepetroules

Sources/SwiftlyCore/FileLock.swift
```diff
@@ -0,0 +1,74 @@
+import Foundation
+import SystemPackage
+
+enum FileLockError: Error {
+    case cannotAcquireLock
```
cmcgee1024, line 5: suggestion (non-blocking): Put a PID as an optional associated value here.

roulpriya: Done @cmcgee1024. Introduced `cannotAcquireLock(FilePath)`, `lockedByPID(FilePath, String)` that provides appropriate error message

Sources/SwiftlyCore/FileLock.swift
```diff
@@ -0,0 +1,109 @@
...
+import SystemPackage
+
+enum FileLockError: Error {
+    case cannotAcquireLock(FilePath)
+    case lockedByPID(FilePath, String?)
+
+    var localizedDescription: String {
+        switch self {
+        case let .cannotAcquireLock(path):
+            return "Cannot acquire lock at \(path). Another process may be holding the lock."
```
cmcgee1024, line 11: suggestion (non-blocking): You can turn this into an actionable error by indicating that the user can remove the file if they are sure that there are no other processes running with the lock.

roulpriya: Done

Sources/SwiftlyCore/ProcessInfo.swift
```diff
@@ -0,0 +1,56 @@
...
+        let error = GetLastError()
+        if error == ERROR_INVALID_PARAMETER || error == ERROR_NOT_FOUND {
+            return false // Process not found
+        } else {
+            throw ProcessCheckError.checkFailed
+        }
+    }
+
+#else
+    throw ProcessCheckError.platformNotSupported
```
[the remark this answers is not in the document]
roulpriya: Done

Sources/SwiftlyCore/FileLock.swift
```diff
@@ -0,0 +1,109 @@
+import Foundation
+import SystemPackage
+
+enum FileLockError: Error {
+    case cannotAcquireLock(FilePath)
+    case lockedByPID(FilePath, String?)
```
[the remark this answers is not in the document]
roulpriya: I have fixed this issue.

cmcgee1024:
@swift-ci test macOS

Sources/SwiftlyCore/FileLock.swift
```diff
@@ -0,0 +1,115 @@
...
+import SystemPackage
+
+enum FileLockError: Error, LocalizedError {
+    case cannotAcquireLock(FilePath)
+    case lockedByPID(FilePath, String)
+
+    var errorDescription: String? {
+        switch self {
+        case let .cannotAcquireLock(path):
+            return "Cannot acquire lock at \(path). Another process may be holding the lock. If you are sure no other processes are running, you can manually remove the lock file at \(path)."
```
cmcgee1024, line 11: nitpick: The path is mentioned twice in the error message when it is probably only needed once.

Sources/SwiftlyCore/FileLock.swift
```diff
@@ -0,0 +1,115 @@
...
+        while Date().timeIntervalSince(start) < timeout {
+            do {
+                return try FileLock(at: path)
+            } catch let FileLockError.lockedByPID(_, pid) {
+                lockedByPID = pid
+                do {
+                    let isRunning = try isProcessRunning(pidString: pid)
+                    if !isRunning {
+                        // Process is no longer running, remove stale lock file and try again
+                        try FileManager.default.removeItem(at: URL(fileURLWithPath: path.string))
```
cmcgee1024, line 67: issue (blocking): I don't think that we should be removing the file ourselves, even if we know the process is local and we have the process ID. The user will need to decide what to do if we exceed the timeout, and that's what the actionable errors are about.

Instead, I think that this section should spin the while loop until either the lock is successfully acquired, or the timeout is exceeded. The error from the last spin of the loop can be thrown to the caller to give them the details of the last attempt.

I think that you might find the Swift Result type useful for capturing either the success or failure of the FileLock init as a simple value, instead of the do/catch. It's up to you if you want to use it. https://developer.apple.com/documentation/swift/result/init(catching:)

Sources/SwiftlyCore/FileLock.swift
```diff
@@ -0,0 +1,115 @@
...
+                        continue
+                    }
+                } catch {
+                    throw FileLockError.cannotAcquireLock(path)
+                }
+                try? await Task.sleep(for: .seconds(pollingInterval) + .milliseconds(Int.random(in: 0...200)))
+            }
+        }
+
+        // Timeout reached, throw an error
```
cmcgee1024, line 77: thought: This check to reconstruct the error may not be needed if you keep the last error, or use the Swift Result I linked above. It's fine as it is though.

Sources/SwiftlyCore/FileLock.swift
roulpriya: I agree with you, I made the necessary changes.

PR #364, opened 2025-05-22
Add macOS job to pull request workflow

cmcgee1024:
Compile and test with macOS Sequoia, and the selected Swift toolchain on
each pull request, and merge to the main branch. Also, check the release
build at the same times. Update the prepare GitHub action script so that
it can install swiftly when requested from macOS.

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -30,10 +30,10 @@ jobs:
     steps:
       - name: Checkout repository
         uses: actions/checkout@v4
-      - name: Activate Swiftly
-        run: source /opt/swiftly/env.sh
+      - name: Prepare the action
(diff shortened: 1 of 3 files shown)


.github/workflows/pull_request.yml
```diff
@@ -22,6 +22,20 @@ jobs:
...
+      fail-fast: false
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v4
+      - name: Prepare the action
+        run: ./scripts/prep-gh-action.sh --install-swiftly
+      - name: Build and Test
+        run: swift test
+        timeout-minutes: 60
+
```
shahmishal, line 38: Is there reason for not using the github-workflow yml file here?

PR #367, opened 2025-05-27
Find symlink managed swiftly when creating proxies

cmcgee1024:
When swiftly is being invoked through a symlink, like what happens with
homebrew, then it should be able to create proxies that point to the swiftly
symlink that is on the path. That symlink is updated whenever swiftly is
upgraded in this environment.

Change the find swiftly binary code to account for this case and return
the absolute path to the swiftly symlink for proxy creation.


cmcgee1024:
@swift-ci test macOS

plemarquand (approved):

PR #369, opened 2025-05-29
Add a new document for choosing toolchains for workflows

cmcgee1024:


Documentation/SwiftlyDocs.docc/swift-version.md
```diff
@@ -0,0 +1,97 @@
...
+//swift-tools-version: 5.10
+
+Package(
+   name: "foo"
+)
+```
+
+From the [SwiftPM documentation](https://docs.swift.org/package-manager/PackageDescription/PackageDescription.html#about-the-swift-tools-version) the first purpose of this version is to declare the version of the Package.swift file's contents so that it can be processed using the correct version of the package description structure. Another purpose is to set the minimum version of the "Swift tools" that are needed to use the package.
+
+From this first purpose, this has not so much to do with the toolchain version, as it does with the need to understand the package description that's inside the file. Newer toolchains will fall back to this version of the description to get the package description. It's also in the best interest of a package to use an older version for this so that downstream packages can continue to use older toolchains.
```
heckj, line 72:
```suggestion
From this first purpose, this has not so much to do with the toolchain version, as it does with the need to understand the package description that's inside the file. Newer toolchains use this version to interpret the package description. It's in the best interest of a package to use an older version so that downstream packages can use older toolchains.
```

Documentation/SwiftlyDocs.docc/swift-version.md
```diff
@@ -0,0 +1,97 @@
...
+
+In terms of the second purpose this acts as a lower-bound of the swift tools for reasons stated above. There are serious tensions between using this as the exact recommended version because there are additional concerns with developing, and testing this package that are not shared by consumers of it, such as features used in testing, formatting, and IDE integration.
```
heckj, line 74:
```suggestion
In terms of the second purpose this acts as a lower-bound of the swift tools for reasons stated above. There are tensions between using this as an exact version because there of concerns with developing and testing that aren't shared by consumers, such as features used in testing, formatting, and IDE integration.
```

Documentation/SwiftlyDocs.docc/swift-version.md
```diff
@@ -0,0 +1,97 @@
...
+
+Not every project that uses a Swift toolchain uses SwiftPM. There are cases where CMake is being used as the build system. In other cases the toolchain is used to access a particular release of clang. Both swiftly and the `.swift-version` file can be used in these cases too.
```
heckj, line 76:
```suggestion
Not every project that uses a Swift toolchain uses SwiftPM. CMake is another common build system. In other cases the toolchain is used to access a particular release of clang. You can use both swiftly and the `.swift-version` file in these cases too.
```

Documentation/SwiftlyDocs.docc/swift-version.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Other versions: Swift Language Version
+
+There is also a version that you can set for the Swift language. For example, SwiftPM allows you to set them for a [package](https://docs.swift.org/package-manager/PackageDescription/PackageDescription.html#swiftversion) in your `Package.swift` file.
```
heckj, line 80:
```suggestion
There is also a version that you can set for the Swift language mode. For example, SwiftPM allows you to set it for a [package](https://developer.apple.com/documentation/packagedescription/swiftlanguagemode) in your `Package.swift` manifest.
```

Documentation/SwiftlyDocs.docc/swift-version.md
```diff
@@ -0,0 +1,97 @@
...
+
+```
+Package(
+    name: "foo",
+    swiftLanguageVersions: [v5, v6]
```
heckj, line 85:
```suggestion
    swiftLanguageMode: [v5, v6]
```

swiftLanguageVersion was created and quickly deprecated

Documentation/SwiftlyDocs.docc/swift-version.md
```diff
@@ -0,0 +1,97 @@
...
+)
+```
+
+This version is used to set the supported language modes in the Swift compiler because it can support handling source code from much older, and newer versions of the language. This has little to do with the recommended toolchain version to use for a project, and is much more about supporting legacy Swift code. There are similar settings for the C, and C++ language standards used by clang.
```
heckj, line 89:
```suggestion
This declares the supported language modes for the Swift compiler, as it supports older and newer versions of the language. This has little to do with the recommended toolchain version, and is more about supporting legacy Swift code. There are similar settings for the C, and C++ language standards used by clang.
```

Documentation/SwiftlyDocs.docc/swift-version.md
```diff
@@ -0,0 +1,97 @@
...
+
+### Verifying a project
+
+When verifying a project in something like a CI system, it's important to pick a toolchain that meets the objective of verifying that the current state of the code with a level of reliability. Using the [test pyramid](https://martinfowler.com/articles/practical-test-pyramid.html) model, you want to get the *majority* of tests running _quickly_, but also *reliably*.
+
+Reliability in toolchain selection is knowing which toolchain has produced good results in the pastat a developer's desk, better would be to use one that's been verified in CI. And so, maybe the CI system uses the toolchain that's documented in the developer's guide? It can be a manual process, such as when the developer's guide is updated. Better would be to have some tooling support that handles both develpers getting started and toolchain selection in CI. In any case swiftly can help install the chosen default toolchain.
```
plemarquand, line 27:
```suggestion
Reliability in toolchain selection is knowing which toolchain has produced good results in the past at a developer's desk, better would be to use one that's been verified in CI. And so, maybe the CI system uses the toolchain that's documented in the developer's guide? It can be a manual process, such as when the developer's guide is updated. Better would be to have some tooling support that handles both develpers getting started and toolchain selection in CI. In any case swiftly can help install the chosen default toolchain.
```

Documentation/SwiftlyDocs.docc/swift-version.md
```diff
@@ -0,0 +1,97 @@
...
+
+It is important to note it is often important to engage in certain kinds of testing that can be slower, and less reliable. The test pyramid model accommodates ths for doing integration, end-to-end testing. These should be fewer in number, and possibly less frequently. Since these tests can be less reliable and slower, sometimes they don't run with every pull request. They are commonly run on a schedule with processes in place to react to failures.
```
plemarquand, line 29: This bit feels off topic from the goal of the document (what toolchain should I use)

Documentation/SwiftlyDocs.docc/swift-version.md
```diff
@@ -0,0 +1,97 @@
...
+```
+Package(
+    name: "foo",
+    swiftLanguageVersions: [v5, v6]
+)
+```
+
+This version is used to set the supported language modes in the Swift compiler because it can support handling source code from much older, and newer versions of the language. This has little to do with the recommended toolchain version to use for a project, and is much more about supporting legacy Swift code. There are similar settings for the C, and C++ language standards used by clang.
+
+Newer toolchains can support much older language versions. For Swift versions, they tend to be only the major version, not the minor, and patch versions that uniquely identify an exact recommended toolchain.
```
heckj, line 91:
```suggestion
```

I think I'd drop this completely - the definition of what language mode is supported is in the docs linked above, and this reads as a repeat of the paragraph above.

Documentation/SwiftlyDocs.docc/swift-version.md
```diff
@@ -0,0 +1,97 @@
...
+
+## Summary
+
+Depending on your workflow, there is often some kind of recommended toolchain version that can help you to succeed in the primary flow. What's needed is a standard way to record this information. The swift version file is a simple format to document the version. Tooling support, including swiftly, can help you to choose this recommended toolchain across different projects. While there are other sources of version information in Swift projects, they are designed for other things.
```
heckj, line 95:
```suggestion
Depending on your workflow, there is often a recommended toolchain version that can help you to succeed. The swift version file is a simple format to document this version, and provides a standard way to record it. Tooling support, including swiftly, helps you to choose this recommended toolchain across different projects.
```

Recommend dropping that final sentence, as I think you want to leave the focus on swift-version and how swiftly uses it, rather than trying to disambiguate from the other versions scattered around.

Documentation/SwiftlyDocs.docc/swift-version.md
```diff
@@ -0,0 +1,97 @@
...
+
+Having a single recommended version doesn't prevent you from using other toolchain versions for various purposes. There is room within the test pyramid model for integration testing with alternate versions. Swiftly has some mechanisms, such as the `swiftly run` version selector syntax, to automate testing with older and newer versions to support this style of integration testing, as well ad-hoc explorations too.
```
heckj, line 97:
```suggestion
Having a single recommended version doesn't prevent you from using other toolchain versions. For example, the test pyramid model highlights integration testing with alternate versions. Swiftly provides the `swiftly run` version selector syntax to automate testing with older and newer versions, as well ad-hoc explorations too.
```

Documentation/SwiftlyDocs.docc/swift-version.md
```diff
@@ -0,0 +1,97 @@
...
+
+It is important to note it is often important to engage in certain kinds of testing that can be slower, and less reliable. The test pyramid model accommodates ths for doing integration, end-to-end testing. These should be fewer in number, and possibly less frequently. Since these tests can be less reliable and slower, sometimes they don't run with every pull request. They are commonly run on a schedule with processes in place to react to failures.
+
+In terms of toolchain selection, it can be important (depending on the project), to have integration testing with different toolchains. Sometimes this involves exploratory testing new Swift toolchain snapshots to signal readiness before the next release arrives. In other cases, a sampling of toolchain versions (often older ones) are used to build the software so that downstream projects using older toolchains can be supported. Testing is harder to achieve since the test cases themselves are not directly consumed by downstream projects, and a project might want to adopt newer test features available in newer toolchains. swiftly can also help to install these toolchains for integration testing with them.
```
plemarquand, line 31:
```suggestion
In terms of toolchain selection, it can be important (depending on the project) to perform integration testing with different toolchains. Sometimes this involves testing nightly Swift toolchain snapshots to signal readiness before the next Swift release arrives. In other cases, a sampling of toolchain versions (often older ones) are used to build the software so that downstream projects using older toolchains can be supported. Testing with older toolchains is harder to achieve since the test cases themselves are not directly consumed by downstream projects, and a project might want to adopt newer test features available in recent toolchains. Swiftly can also help to install these older toolchains for integration testing with them.
```

Documentation/SwiftlyDocs.docc/swift-version.md
```diff
@@ -0,0 +1,97 @@
...
+
+```
+swiftly run +6.2-snapshot make
```
plemarquand, line 34: What are these commands for?

Documentation/SwiftlyDocs.docc/swift-version.md
plemarquand, line 95:
```suggestion
Depending on your workflow, there is often a recommended toolchain version that will successfully build the project. The `.swift-version` acts as a simple format to document this toolchain version. Tooling support, including swiftly, can help you to choose this recommended toolchain across different projects. While there are other sources of version information in Swift projects, they are designed for other things.
```

PR #370, opened 2025-05-31
Add an option to self-update to provide the swiftly version to update

cmcgee1024:
This can be useful for testing the self-update to pre-releases that may not
yet be published as the current in the swift.org API because they are undergoing
testing.

### DIFF
--- a/Documentation/SwiftlyDocs.docc/swiftly-cli-reference.md
+++ b/Documentation/SwiftlyDocs.docc/swiftly-cli-reference.md
@@ -464,7 +464,7 @@ swiftly init [--no-modify-profile] [--overwrite] [--platform=<platform>] [--skip
 Update the version of swiftly itself.
 
 ```
-swiftly self-update [--assume-yes] [--verbose] [--version] [--help]
+swiftly self-update [--assume-yes] [--verbose]  [--version] [--help]
 ```
 
 **--assume-yes:**
--- a/Documentation/SwiftlyDocs.docc/swiftly-cli-reference.md
+++ b/Documentation/SwiftlyDocs.docc/swiftly-cli-reference.md
@@ -152,7 +152,7 @@ Note that listing available snapshots before the latest release (major and minor
 Set the in-use or default toolchain. If no toolchain is provided, print the currently in-use toolchain, if any.
 
 ```
-swiftly use [--print-location] [--global-default] [--assume-yes] [--verbose] [<toolchain>] [--version] [--help]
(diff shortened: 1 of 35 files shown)


Documentation/SwiftlyDocs.docc/swiftly-cli-reference.md
```diff
@@ -464,7 +464,7 @@ swiftly init [--no-modify-profile] [--overwrite] [--platform=<platform>] [--skip
 Update the version of swiftly itself.
 
 ```
-swiftly self-update [--assume-yes] [--verbose] [--version] [--help]
+swiftly self-update [--assume-yes] [--verbose]  [--version] [--help]
```
etcwilde, line 467: Extra space added here

Sources/Swiftly/SelfUpdate.swift
```diff
@@ -31,50 +39,76 @@ struct SelfUpdate: SwiftlyCommand {
...
 
-        guard try swiftlyRelease.swiftlyVersion > SwiftlyCore.version else {
-            await ctx.print("Already up to date.")
-            return SwiftlyCore.version
+#if os(Linux)
+#if arch(x86_64)
+            downloadURL = URL(string: "https://download.swift.org/swiftly/linux/swiftly-\(version)-x86_64.tar.gz")
+#elseif arch(arm64)
+            downloadURL = URL(string: "https://download.swift.org/swiftly/linux/swiftly-\(version)-aarch64.tar.gz")
+#endif
```
etcwilde, line 61: Should this throw an error if we somehow get here and don't know what we're downloading?
As it stands, I think this will end up running the update based on `getCurrentSwiftlyRelease()` since `downloadURL` won't get set, even though a version was specified.

Tests/SwiftlyTests/SelfUpdateTests.swift
```diff
@@ -21,7 +21,7 @@ import Testing
     func runSelfUpdateTest(latestVersion: SwiftlyVersion) async throws {
         try await SwiftlyTests.withTestHome {
             try await SwiftlyTests.withMockedSwiftlyVersion(latestSwiftlyVersion: latestVersion) {
-                let updatedVersion = try await SelfUpdate.execute(SwiftlyTests.ctx, verbose: true)
+                let updatedVersion = try await SelfUpdate.execute(SwiftlyTests.ctx, verbose: true, version: nil)
```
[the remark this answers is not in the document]
cmcgee1024: Thanks, I've added more test coverage for this

PR #381, opened 2025-06-12
Replace hardcoded config fallback version with dynamic constant

belkhadir:
Previously, when loading the config file, the fallback version `"0.3.0"` was hardcoded to handle unversioned configs from earlier releases.

This MR replaces the hardcoded value with the centralized `SwiftlyCore.version` constant, ensuring consistency across the codebase and reducing the risk of version mismatches in the future.

### DIFF
--- a/Sources/Swiftly/Init.swift
+++ b/Sources/Swiftly/Init.swift
@@ -179,7 +179,7 @@ struct Init: SwiftlyCommand {
         if overwrite || config == nil {
             let pd = try await Swiftly.currentPlatform.detectPlatform(ctx, disableConfirmation: assumeYes, platform: platform)
             let c = Config(inUse: nil, installedToolchains: [], platform: pd, version: SwiftlyCore.version)
-            
+
             try c.save(ctx)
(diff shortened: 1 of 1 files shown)


cmcgee1024:
Thank you for this contribution.

The code is working as designed at the time, which is that the version would have to be "0.3.0" (or earlier) if there's no version field in the config.json. This sets the version so that it can be compared further down the line for any migrations of the swiftly installation.

Since swiftly does not support upgrading from "0.3.0" and earlier the change here might be to make the version field mandatory, and remove this set if it's not set code.

belkhadir:
Thanks for the clarification; that makes perfect sense.

I've updated the code to make the `version` field in the config mandatory and removed the fallback to `"0.3.0"`. The initializer now requires a version, and all optional checks have been cleaned up.

cmcgee1024:
@belkhadir there's a few compile errors in the tests, and some code reformatting needed for this. To reformat the code you can use `swift run swiftformat Sources Tests`

Issue #376, opened 2025-06-10
Add JSON Output Format


roulpriya:
Add support for JSON output for toolchain management commands:
- List toolchains
- Install toolchain
- Use toolchain

Create structured result types for each command operation, define clear structs/classes for outputs like toolchain lists, version info, etc and include fields for all data currently output as text

Modify command implementations to return these structured types instead of directly printing text, and separate logic from presentation concerns

Implement a rendering layer with text and JSON formatters by creating interfaces with implementations for each output format. 

Use a command-line argument like `--format=json` to control the output format

For backward compatibility:
- Default to text output when no format is specified
- Ensure text formatting matches the current output closely

PR #387, opened 2025-06-17
Refactor list-available command to support json format

roulpriya:
- Fixes Issue #376 
- Description
  Refactor `list-available` command to support JSON output format

### DIFF
--- a/Sources/SwiftlyCore/OutputFormatter.swift
+++ b/Sources/SwiftlyCore/OutputFormatter.swift
@@ -11,10 +11,10 @@ public enum OutputFormat: String, Sendable, CaseIterable, ExpressibleByArgument 
 }
 
 public protocol OutputFormatter {
-    func format(_ data: OutputData) -> String
+    func format(_ data: OutputData) throws -> String
 }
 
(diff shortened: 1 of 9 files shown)


Sources/SwiftlyCore/OutputFormatter.swift
```diff
@@ -26,19 +26,21 @@ public struct TextOutputFormatter: OutputFormatter {
...
         let encoder = JSONEncoder()
         encoder.outputFormatting = [.prettyPrinted, .sortedKeys]
 
-        let jsonData = try? encoder.encode(data)
-
-        guard let jsonData = jsonData, let result = String(data: jsonData, encoding: .utf8) else {
-            return "{}"
+        let jsonData = try encoder.encode(data)
+        guard let result = String(data: jsonData, encoding: .utf8) else {
+            throw OutputFormatterError.encodingError("Failed to convert JSON data to string.")
```
cmcgee1024, line 42: suggestion: This message is in relation to converting byte Data to a String using the UTF8 encoding. This could be reworded a bit to something that makes this a bit more clear in the very rare case where it occurs. How about something like "failed to encode JSON data as a string in UTF-8"?

Issue #254, opened 2025-03-19
Add a progress-file option to swiftly install and swiftly init


cmcgee1024:
It is possible that a GUI could be driving swiftly in these two cases where things can take a significant time, mostly the downloading the toolchain portion. Add some kind of progress-file option to swiftly so that it can report back progress periodically to the GUI.

This option can be just a simple path, provided by the client, and swiftly appends JSON Lines of progress output there. Note that `swift test` has a similar capability that's currently used by VSCode to report test execution progress.

It will be up to the client to pre-create, or otherwise make available the file. It could be a regular file, or a named pipe, or even a device (e.g. /dev/stdout on Unix). The client then reads or tails from this file and updates the UI with the information.

See this module in swift testing for an idea of the JSON Line records it uses for this purpose: https://github.com/swiftlang/swift-testing/tree/main/Sources/Testing/ABI

cmcgee1024:
Consider versioning with this option too (e.g. `--progress-file-version=0`. It could be that the schema changes over time.

louisunlimited:
I do realize this is a part of GSoC 2025 for integrating swiftly with VSCode (and I did submit a proposal for it), but since this is a public issue I went ahead and made a little poc for `swiftly install`, hope this helps anybody eventually picking up the project, and also possibly bring up some discussion around it.

- #343

PR #391, opened 2025-06-21
Json Progress File

roulpriya:
Fixes Issue #254 

- Added a JsonFileProgressReporter that writes progress updates and completion status as JSON lines to a file during toolchain installation.
- Integrated with `swiftly install`, allowing progress to be tracked.
- Implemented and tested progress reporting, including step updates, completion (success/failure), and file clearing.

### DIFF
--- a/Documentation/SwiftlyDocs.docc/swiftly-cli-reference.md
+++ b/Documentation/SwiftlyDocs.docc/swiftly-cli-reference.md
@@ -86,6 +86,7 @@ written to this file as commands that can be run after the installation.
 
 Progress information will be appended to this file as JSON objects, one per line.
 Each progress entry contains timestamp, progress percentage, and a descriptive message.
+The file must be writable, else an error will be thrown.
 
 
 **--assume-yes:**
--- a/Documentation/SwiftlyDocs.docc/swiftly-cli-reference.md
+++ b/Documentation/SwiftlyDocs.docc/swiftly-cli-reference.md
@@ -23,7 +23,7 @@ swiftly [--version] [--help]
 Install a new toolchain.
 
 ```
-swiftly install [<version>] [--use] [--verify|no-verify] [--post-install-file=<post-install-file>] [--assume-yes] [--verbose] [--version] [--help]
+swiftly install [<version>] [--use] [--verify|no-verify] [--post-install-file=<post-install-file>] [--progress-file=<progress-file>] [--assume-yes] [--verbose] [--version] [--help]
 ```
 
 **version:**
@@ -78,6 +78,14 @@ NOTE: Swiftly downloads toolchains to a temporary file that it later cleans duri
 
 If the toolchain that is installed has extra post installation steps, they will be
 written to this file as commands that can be run after the installation.
+
+
+**--progress-file=\<progress-file\>:**
+
+*A file path where progress information will be written in JSONL format*
+
+Progress information will be appended to this file as JSON objects, one per line.
+Each progress entry contains timestamp, progress percentage, and a descriptive message.
 
 
 **--assume-yes:**
--- a/Sources/Swiftly/JsonFileProgressReporter.swift
+++ b/Sources/Swiftly/JsonFileProgressReporter.swift
@@ -11,27 +11,30 @@ enum ProgressInfo: Codable {
 struct JsonFileProgressReporter: ProgressAnimationProtocol {
     let filePath: FilePath
     private let encoder: JSONEncoder
+    private let ctx: SwiftlyCoreContext
+    private let fileHandle: FileHandle
 
-    init(filePath: FilePath, encoder: JSONEncoder = JSONEncoder()) {
+    init(_ ctx: SwiftlyCoreContext, filePath: FilePath, encoder: JSONEncoder = JSONEncoder()) throws
+    {
+        self.ctx = ctx
         self.filePath = filePath
         self.encoder = encoder
+        self.fileHandle = try FileHandle(forWritingTo: URL(fileURLWithPath: filePath.string))
     }
 
     private func writeProgress(_ progress: ProgressInfo) {
         let jsonData = try? self.encoder.encode(progress)
-        guard let jsonData = jsonData, let jsonString = String(data: jsonData, encoding: .utf8)
-        else {
-            print("Failed to encode progress entry to JSON")
+        guard let jsonData = jsonData else {
+            Task { [ctx = self.ctx] in
+                await ctx.message("Failed to encode progress entry to JSON")
+            }
             return
         }
 
-        let jsonLine = jsonString + "\n"
-
-        do {
-            try jsonLine.append(to: self.filePath)
-        } catch {
-            print("Failed to write progress entry to \(self.filePath): \(error)")
-        }
+        self.fileHandle.seekToEndOfFile()
+        self.fileHandle.write(jsonData)
+        self.fileHandle.write("\n".data(using: .utf8) ?? Data())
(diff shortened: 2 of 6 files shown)


Sources/SwiftlyCore/FileManager+FilePath.swift
```diff
@@ -190,6 +190,22 @@ extension String {
...
     }
 
+    public func append(to path: FilePath, encoding enc: String.Encoding = .utf8) throws {
+        if !FileManager.default.fileExists(atPath: path.string) {
+            try self.write(to: path, atomically: true, encoding: enc)
+            return
+        }
+
+        let fileHandle = try FileHandle(forWritingTo: URL(fileURLWithPath: path.string))
+        defer { fileHandle.closeFile() }
```
cmcgee1024, line 200: praise: This is exactly what I hope for with the defer capability in Swift. When I see something allocated, the release logic follows it, answering my question of "how does this get deallocated?"

Sources/SwiftlyCore/FileManager+FilePath.swift
```diff
@@ -190,6 +190,22 @@ extension String {
         try self.write(to: URL(fileURLWithPath: path.string), atomically: atomically, encoding: enc)
     }
 
+    public func append(to path: FilePath, encoding enc: String.Encoding = .utf8) throws {
+        if !FileManager.default.fileExists(atPath: path.string) {
+            try self.write(to: path, atomically: true, encoding: enc)
```
cmcgee1024, line 195: suggestion: It might be worth a comment here that in theory there is a race between detecting if the file exists and the atomic write to it. swiftly operates on the main actor (for now), and so this is not usually a problem in practice.

Sources/SwiftlyCore/FileManager+FilePath.swift
```diff
@@ -190,6 +190,22 @@ extension String {
         try self.write(to: URL(fileURLWithPath: path.string), atomically: atomically, encoding: enc)
     }
 
+    public func append(to path: FilePath, encoding enc: String.Encoding = .utf8) throws {
+        if !FileManager.default.fileExists(atPath: path.string) {
```
cmcgee1024, line 194: issue: The progress file lifecycle (and type) is managed by the client of swiftly, so there should be no need to detect if it is exists here. Just go ahead and append to it, and fail if the append fails. This allows the client to allocate a "named pipe" so that they can get the progress streamed directly to a file descriptor that it monitors so that it can get immediate progress monitor updates.

Documentation/SwiftlyDocs.docc/swiftly-cli-reference.md
```diff
@@ -80,6 +80,14 @@ If the toolchain that is installed has extra post installation steps, they will
 written to this file as commands that can be run after the installation.
 
 
+**--progress-file=\<progress-file\>:**
+
+*A file path where progress information will be written in JSONL format*
+
+Progress information will be appended to this file as JSON objects, one per line.
```
cmcgee1024, line 87: suggestion: Let's spell out that the contract that the client must create this file first, and swiftly will append to it as it progresses.

Sources/SwiftlyCore/FileManager+FilePath.swift
```diff
@@ -190,6 +190,22 @@ extension String {
...
+            try self.write(to: path, atomically: true, encoding: enc)
+            return
+        }
+
+        let fileHandle = try FileHandle(forWritingTo: URL(fileURLWithPath: path.string))
+        defer { fileHandle.closeFile() }
+        fileHandle.seekToEndOfFile()
```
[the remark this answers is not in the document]
cmcgee1024: You might find that you need to `fflush()` each line. Hopefully there's a Swift way of doing this instead of having to make the sys call.

Tests/SwiftlyTests/InstallTests.swift
```diff
@@ -262,4 +263,37 @@ import Testing
...
+
+        #expect(try await fs.exists(atPath: progressFile))
+
+        let progressContent = try String(contentsOfFile: progressFile.string)
+        let lines = progressContent.components(separatedBy: .newlines).filter { !$0.isEmpty }
+
+        #expect(!lines.isEmpty, "Progress file should contain progress entries")
+
+        // Verify that at least one progress entry exists
+        let hasProgressEntry = lines.contains { line in
```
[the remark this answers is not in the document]
roulpriya: changed all the tests to do this now

Sources/Swiftly/JsonFileProgressReporter.swift
```diff
@@ -0,0 +1,62 @@
...
+    private func writeProgress(_ progress: ProgressInfo) {
+        let jsonData = try? self.encoder.encode(progress)
+        guard let jsonData = jsonData else {
+            Task { [ctx = self.ctx] in
+                await ctx.message("Failed to encode progress entry to JSON")
+            }
+            return
+        }
+
+        self.fileHandle.seekToEndOfFile()
```
cmcgee1024, line 34: suggestion: I think that seeking shouldn't be required here since we have a long-lived file handle, and the expectation is that there's only one writer.

Tests/SwiftlyTests/JsonFileProgressReporterTests.swift
```diff
@@ -0,0 +1,163 @@
+import Foundation
+import SystemPackage
+import Testing
+
+@testable import Swiftly
+@testable import SwiftlyCore
+
+@Suite struct JsonFileProgressReporterTests {
+    @Test("Test update method writes progress to file as valid JSONNL")
+    func testUpdateWritesProgressToFile() async throws {
```
cmcgee1024, line 10: question: Can there be a test that creates a named pipe, and tests that information can be streamed in real time? If this is too difficult to make a good reliable test, then please ignore this.

Sources/Swiftly/JsonFileProgressReporter.swift
```diff
@@ -0,0 +1,62 @@
...
+                percent: Int(Double(step) / Double(total) * 100),
+                text: text
+            ))
+    }
+
+    func complete(success: Bool) {
+        self.writeProgress(ProgressInfo.complete(success: success))
+    }
+
+    func clear() {
```
cmcgee1024, line 54: question: Where's the requirement for a progress file clear action coming from? I feel that this isn't a necessary function, and is likely to fail on a pipe instead of a regular file.

roulpriya: It is part of the `ProgressAnimationProtocol`. Do you think we can keep it unimplemented then?

cmcgee1024: Yes, I think that a no-op makes sense here. The next progress entry will probably go backwards, and it will be up to a client to handle that case.

Tests/SwiftlyTests/InstallTests.swift
```diff
@@ -262,4 +263,134 @@ import Testing
...
+                "--progress-file=\(pipePath)",
+            ])
+        }
+
+        await withTaskGroup(of: Void.self) { group in
+            group.addTask { try? await readerTask.value }
+            group.addTask { try? await installTask.value }
+        }
+
+        try await Task.sleep(nanoseconds: 100_000_000) // 100ms
```
cmcgee1024, line 373: question: Is this sleep required? I think that awaiting the `withTaskGroup` above should wait until both child tasks are done.

roulpriya: I was not sure, hence added sleep, I removed it.

PR #394, opened 2025-06-25
Add progress reporter protocol

roulpriya:
This PR contains
- Create a ProgressReporterProtocol to support an async function.

### DIFF
--- a/Sources/Swiftly/Install.swift
+++ b/Sources/Swiftly/Install.swift
@@ -350,22 +350,28 @@ struct Install: SwiftlyCommand {
 
                         lastUpdate = Date()
 
-                        await animation.update(
-                            step: progress.receivedBytes,
-                            total: progress.totalBytes!,
-                            text:
-                            "Downloaded \(String(format: "%.1f", downloadedMiB)) MiB of \(String(format: "%.1f", totalMiB)) MiB"
-                        )
+                        do {
+                            try await animation.update(
+                                step: progress.receivedBytes,
+                                total: progress.totalBytes!,
+                                text:
+                                "Downloaded \(String(format: "%.1f", downloadedMiB)) MiB of \(String(format: "%.1f", totalMiB)) MiB"
+                            )
+                        } catch {
+                            await ctx.message(
+                                "Failed to update progress: \(error.localizedDescription)"
+                            )
+                        }
                     }
                 )
             } catch let notFound as DownloadNotFoundError {
                 throw SwiftlyError(
                     message: "\(version) does not exist at URL \(notFound.url), exiting")
             } catch {
-                await animation.complete(success: false)
+                try? await animation.complete(success: false)
                 throw error
             }
-            await animation.complete(success: true)
+            try await animation.complete(success: true)
 
             if verifySignature {
                 try await Swiftly.currentPlatform.verifyToolchainSignature(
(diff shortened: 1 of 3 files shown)


Sources/Swiftly/ProgressReporter.swift
```diff
@@ -0,0 +1,90 @@
+import Foundation
+import SwiftlyCore
+import SystemPackage
+import TSCBasic
+import TSCUtility
+
+public protocol ProgressReporterProtocol {
+    /// Updates the progress animation with the current step, total steps, and an optional text message.
+    func update(step: Int, total: Int, text: String) async
```
cmcgee1024, line 9: suggestion: Since update can be fallible due to one of the implementations doing I/O operations, this can be declared as `throws` and then the `try?` become plain `try` and the original I/O error can flow all of the way to the top-level error reporting.

Sources/Swiftly/ProgressReporter.swift
```diff
@@ -0,0 +1,90 @@
...
+
+    /// Completes the progress animation, indicating success or failure.
+    func complete(success: Bool) async
```
cmcgee1024, line 12: suggestion: Same as above, this can be declared `throws` so that I/O errors reported on the complete can be thrown to an error reported with the original I/O error details.

Sources/Swiftly/ProgressReporter.swift
```diff
@@ -0,0 +1,90 @@
...
+    init(_ ctx: SwiftlyCoreContext, filePath: FilePath, encoder: JSONEncoder = JSONEncoder()) throws
+    {
+        self.ctx = ctx
+        self.filePath = filePath
+        self.encoder = encoder
+        self.fileHandle = try FileHandle(forWritingTo: URL(fileURLWithPath: filePath.string))
+    }
+
+    private func writeProgress(_ progress: ProgressInfo) async {
+        let jsonData = try? self.encoder.encode(progress)
```
cmcgee1024, line 59: suggestion: From above, if this method is declared `throws` then the errors can be thrown directly to the caller with the particular details of the failed I/O operation.

Sources/Swiftly/ProgressReporter.swift
```diff
@@ -0,0 +1,90 @@
...
+        guard let jsonData = jsonData else {
+            await self.ctx.message("Failed to encode progress entry to JSON")
```
cmcgee1024, line 61: praise: It's great to see this message being reliably reported now that the method is async.

Sources/Swiftly/Install.swift
```diff
@@ -351,22 +350,26 @@ struct Install: SwiftlyCommand {
...
+                            // Silently ignore progress update errors to avoid interrupting download
+                        }
                     }
                 )
             } catch let notFound as DownloadNotFoundError {
                 throw SwiftlyError(
                     message: "\(version) does not exist at URL \(notFound.url), exiting")
             } catch {
-                animation.complete(success: false)
+                try? await animation.complete(success: false)
```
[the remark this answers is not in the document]
cmcgee1024: @roulpriya in this catch block, there's an error that was thrown. The `animation.complete()` can also throw an error. I think that the first error is probably more important and should be the one that is thrown. This code is doing the right thing in my opinion. :+1

PR #403, opened 2025-07-21
[WIP] Add list-dependencies command

justice-adams-apple:
Follow up from https://github.com/swiftlang/swiftly/pull/350 and https://github.com/swiftlang/swiftly/pull/352
* Add a new command `list-depenencies` which lists toolchain dependencies for the given platorm
* Ask to install the dependencies for the user with elevated privileges and give option to auto install with `-I`

# ToDo: Add tests

### DIFF
--- a/Sources/Swiftly/ListDependencies.swift
+++ b/Sources/Swiftly/ListDependencies.swift
@@ -10,6 +10,9 @@ struct ListDependencies: SwiftlyCommand {
     @Option(name: .long, help: "Output format (text, json)")
     var format: SwiftlyCore.OutputFormat = .text
 
+    @Flag(name: .shortAndLong, help: "Automatically install missing system dependencies with elevated permissions")
+    var installSystemDeps: Bool = false
+
     internal static var allowedInstallCommands: Regex<(Substring, Substring, Substring)> { try! Regex("^(apt-get|yum) -y install( [A-Za-z0-9:\\-\\+]+)+$") }
 
     mutating func run() async throws {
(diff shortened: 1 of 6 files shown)


Sources/Swiftly/ListDependencies.swift
```diff
@@ -0,0 +1,98 @@
...
+struct ListDependencies: SwiftlyCommand {
+    public static let configuration = CommandConfiguration(
+        abstract: "List toolchain dependencies required for the given platform."
+    )
+
+    @Option(name: .long, help: "Output format (text, json)")
+    var format: SwiftlyCore.OutputFormat = .text
+
+    @Flag(name: .shortAndLong, help: "Automatically install missing system dependencies with elevated permissions")
+    var installSystemDeps: Bool = false
```
cmcgee1024, line 14: issue: Let's call out sudo as the mechanism that will be used to install the system deps here to make that part clear. Sudo may or may not work for some users and machines.

Sources/Swiftly/ListDependencies.swift
```diff
@@ -0,0 +1,98 @@
...
+            // This is very security sensitive code here and that's why there's special process handling
+            // and an allow-list of what we will attempt to run as root. Also, the sudo binary is run directly
+            // with a fully-qualified path without any checking in order to avoid TOCTOU.
+            guard try Self.allowedInstallCommands.wholeMatch(in: installCmd) != nil else {
+                fatalError("Command \(installCmd) does not match allowed patterns for sudo")
+            }
+
+            let p = Process()
+            p.executableURL = URL(fileURLWithPath: "/usr/bin/sudo")
+            p.arguments = ["-k"] + ["-p", "Enter your sudo password to run the dependency install command right away (Ctrl-C aborts): "] + installCmd.split(separator: " ").map { String($0) }
```
cmcgee1024, line 78: thought: This pattern of invoking sudo came from my previous attempt to eliminate prompts where they weren't needed. It might be better to remove the "-k" and "-p" flags here, and just include the details in the above prompts.

PR #410, opened 2025-08-13
Use Subprocess for process management

cmcgee1024:


Sources/SwiftlyCore/Platform.swift
```diff
@@ -254,39 +260,65 @@ extension Platform {
...
+                arguments: .init(args),
+                environment: env != nil ? .inherit.updating(env ?? [:]) : .inherit,
+                output: .fileDescriptor(.standardError, closeAfterSpawningProcess: false),
+                error: .fileDescriptor(.standardError, closeAfterSpawningProcess: false),
+            )
 
-        if let env {
-            process.environment = env
-        }
+            // TODO: figure out how to set the process group
```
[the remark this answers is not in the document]
cmcgee1024: This was here because Foundation Process would automatically assign a new process group to child processes, and we wanted to be able to have the terminal control the child process that is in the foreground. With the default behaviour of swift-subprocess being that the child is in the current process's group this is no longer needed, which is a nice bonus of this new API.

Sources/TestSwiftly/TestSwiftly.swift
```diff
@@ -92,55 +97,58 @@ struct TestSwiftly: AsyncParsableCommand {
...
 
-            try currentPlatform.runProgram(extractedSwiftly.string, "init", "--assume-yes", "--no-modify-profile", "--skip-install", quiet: false, env: env)
-            try await sh(executable: .path(shell), .login, .command(". \"\(customLoc! / "env.sh")\" && swiftly install --assume-yes latest --post-install-file=./post-install.sh")).run(currentPlatform, env: env, quiet: false)
+            env = env.updating([
+                "SWIFTLY_HOME_DIR": customLoc!.string,
+                "SWIFTLY_BIN_DIR": (customLoc! / "bin").string,
+                "SWIFTLY_TOOLCHAINS_DIR": (customLoc! / "toolchains").string,
+            ])
+
+            _ = try await Subprocess.run(.path(extractedSwiftly), arguments: ["init", "--assume-yes", "--no-modify-profile", "--skip-install"], environment: env, input: .standardInput, output: .standardOutput, error: .standardError)
```
cmcgee1024: Revisit these to make sure that the exit codes are capture correctly and errors are thrown when there are failures.

Sources/TestSwiftly/TestSwiftly.swift
```diff
@@ -92,55 +97,58 @@ struct TestSwiftly: AsyncParsableCommand {
...
+                env = env.updating(["ZDOTDIR": fs.home.string])
             } else if shell.ends(with: "fish") {
-                env["XDG_CONFIG_HOME"] = (fs.home / ".config").string
+                env = env.updating(["XDG_CONFIG_HOME": (fs.home / ".config").string])
             }
 
-            try currentPlatform.runProgram(extractedSwiftly.string, "init", "--assume-yes", "--skip-install", quiet: false, env: env)
-            try await sh(executable: .path(shell), .login, .command("swiftly install --assume-yes latest --post-install-file=./post-install.sh")).run(currentPlatform, env: env, quiet: false)
+            _ = try await Subprocess.run(.path(extractedSwiftly), arguments: ["init", "--assume-yes", "--skip-install"], environment: env, input: .standardInput, output: .standardOutput, error: .standardError)
+            _ = try await sh(executable: .path(shell), .login, .command("swiftly install --assume-yes latest --post-install-file=./post-install.sh")).run(environment: env, output: .standardOutput, error: .standardError)
```
cmcgee1024: Capturing the return value shouldn't be necessary here with the run method since it will throw on non-success case. Try using the variant with optional quiet parameter.

jakepetroules (approved):

PR #411, opened 2025-08-22
Fix Zsh auto-completions setup instructions

tomassliz:
Use the plural form for the "completions" folder to maintain consistent naming.

Prefix the generated Zsh completions file with an underscore; otherwise, it won't match the expected naming convention, and the completions will not work.


etcwilde (approved):

samuelmurray:
It's unfortunate that https://www.swift.org/swiftly/documentation/swiftly/shell-autocompletion is not updated with these changes. Does updating the website require some manual trigger? @etcwilde

PR #419, opened 2025-09-16
Detect HTTPS_PROXY in HTTPClient

justice-adams-apple:
Our build jobs won't work unless we detect and add the https proxy into the HTTP client prior to downloading the lib archive dependency. We can reuse the `HTTPRequestExecutorImpl` to detect and use the proxy correctly

### DIFF
--- a/Tools/build-swiftly-release/BuildSwiftlyRelease.swift
+++ b/Tools/build-swiftly-release/BuildSwiftlyRelease.swift
@@ -150,8 +150,9 @@ struct BuildSwiftlyRelease: AsyncParsableCommand {
         try? await fs.remove(atPath: libArchivePath)
 
         // Download libarchive
+        let httpExecutor = HTTPRequestExecutorImpl()
         let libarchiveRequest = HTTPClientRequest(url: "https://github.com/libarchive/libarchive/releases/download/v\(libArchiveVersion)/libarchive-\(libArchiveVersion).tar.gz")
-        let libarchiveResponse = try await HTTPClient.shared.execute(libarchiveRequest, timeout: .seconds(60))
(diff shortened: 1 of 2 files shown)


Tools/build-swiftly-release/BuildSwiftlyRelease.swift
```diff
@@ -150,8 +150,15 @@ struct BuildSwiftlyRelease: AsyncParsableCommand {
         try? await fs.remove(atPath: libArchivePath)
 
         // Download libarchive
+        let httpExecutor = HTTPRequestExecutorImpl()
+        defer {
+            Task {
```
cmcgee1024, line 155: suggestion: If you leverage the deinit of the httpExecutor then this defer shouldn't be needed here. If an explicit shutdown is needed then it should probably use the syncShutdown() of the http client so that this defer doesn't need to launch an unstructured task that may or may not be run when expected since it is asynchronous.

Tools/build-swiftly-release/BuildSwiftlyRelease.swift
```diff
@@ -150,8 +150,15 @@ struct BuildSwiftlyRelease: AsyncParsableCommand {
...
+                try? await httpExecutor.shutdown()
+            }
+        }
+
         let libarchiveRequest = HTTPClientRequest(url: "https://github.com/libarchive/libarchive/releases/download/v\(libArchiveVersion)/libarchive-\(libArchiveVersion).tar.gz")
-        let libarchiveResponse = try await HTTPClient.shared.execute(libarchiveRequest, timeout: .seconds(60))
```
cmcgee1024, line 154: praise: this looks good for cases where the build script is run in a http proxy environment.

thought: Maybe we should have a workflow that verifies that this build script works in such an environment. I'm not quite sure how to configure Linux so that the network traffic will be blocked unless it goes through a proxy.

PR #422, opened 2025-09-19
Update the automated install guide to cover private macOS install

cmcgee1024:
There is some additional information needed for fully private macOS swiftly
installations. Add this information to the automated installation guide.

### DIFF
--- a/Documentation/SwiftlyDocs.docc/automated-install.md
+++ b/Documentation/SwiftlyDocs.docc/automated-install.md
@@ -9,18 +9,28 @@ The examples are based on a typical Unix environment.
 ### Download the binary
 
 First, download the swiftly binary from swift.org for your operating system (for example, Linux) and processor architecture (for example `arm64` or `x86_64`).
-Here's an example using the popular curl command.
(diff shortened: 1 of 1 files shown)


Documentation/SwiftlyDocs.docc/automated-install.md
```diff
@@ -9,20 +9,30 @@ The examples are based on a typical Unix environment.
...
 
-On macOS, download the pkg file and extract it like this from the command-line:
+On macOS, you can download the pkg file and extract it like this from the command-line:
 
 ```
 curl -L https://download.swift.org/swiftly/darwin/swiftly.pkg > swiftly.pkg
 installer -pkg swiftly.pkg -target CurrentUserHomeDirectory
 ```
 
+You could be installing swiftly on a shared user account for macOS. In that case you might want to extract the pkg in a private location where it cannot be affected by other processes running in the same user account.
```
heckj, line 28: This would be easier to parse if you laid this out directly:


```suggestion
If you are installing swiftly on a shared user account for macOS, extract the `pkg` in a private location where it cannot be affected by other processes running in the same user account.
```

PR #446, opened 2025-10-19
feat: add support for Gentoo Linux in HTTP client and package management

coleleavitt:
This pull request adds support for Gentoo Linux as a recognized platform in the codebase. It introduces Gentoo-specific configuration, package lists, and package manager handling, ensuring that Gentoo can be properly detected and managed alongside other Linux distributions.

**Gentoo Linux support:**

* Added `gentoo` to the list of known Linux platforms in `PlatformDefinition` and updated platform detection logic to recognize Gentoo systems. [[1]](diffhunk://#diff-b34b6716eb8e703f78923b67daf2df4ac1d56b5bb35500fca27ed1fe47c2cda0R49-R51) [[2]](diffhunk://#diff-8ad2e133469c216218c352dfe9235fabd7bc76e415aba45a538c12dfc997d0e5R615-R616)
* Included Gentoo in the list of platforms handled by the HTTP client, enabling API compatibility.

**Package management and configuration:**

* Added Gentoo-specific package list and mapped the package manager to `emerge` in the `Linux` platform implementation. [[1]](diffhunk://#diff-8ad2e133469c216218c352dfe9235fabd7bc76e415aba45a538c12dfc997d0e5R21) [[2]](diffhunk://#diff-8ad2e133469c216218c352dfe9235fabd7bc76e415aba45a538c12dfc997d0e5R231-R249) [[3]](diffhunk://#diff-8ad2e133469c216218c352dfe9235fabd7bc76e415aba45a538c12dfc997d0e5R273-R274)
* Implemented package installation checks using Portage's `qlist` for Gentoo.

### DIFF
--- a/Sources/LinuxPlatform/Linux.swift
+++ b/Sources/LinuxPlatform/Linux.swift
@@ -18,6 +18,7 @@ public struct Linux: Platform {
         .rhel9,
         .amazonlinux2,
         .debian12,
+        .gentoo,
     ]
 
     public init() {}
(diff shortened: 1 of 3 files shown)


coleleavitt:
> Gentoo isn't a supported platform for the Swift toolchain and swiftly doesn't yet have support for explicitly marking platforms as experimental. This change would need to make it clear that Gentoo is in this experimental category and that it should be used at one's own risk.
> 
> There should be some kind of verification of Gentoo as well, so there would need to be a workflow added for it. This shouldn't be too difficult to add based on an official docker hub label. However, this workflow should be non-voting in case it breaks at some point.

gentoo actually does have a package for swift through the guru repository and also I was able to override the installation of swiftly for gentoo using the tool by just spoofing the version to ubuntu.

https://github.com/gentoo/guru/tree/master/dev-lang/swift
https://bugs.gentoo.org/567514

Issue #14, opened 2022-10-08
Add `--quiet` option to supress progress output


patrickfreed:
swiftly currently produces progress output when downloading to give the user an idea of how long an install will take. This is ideal for humans using it via the CLI, but it can be pretty noisy when included in the logs of a CI test run. We should provide a `--quiet` argument that suppresses this output for such uses cases. Error and possibly success output would still be produced.

We probably would want to enable this for swiftly's tests too.

khushi1711:
Hi @patrickfreed can I take this issue?

PR #472, opened 2025-12-01
Added quiet flag to hide progress animation during install

asmeili:
Adds a new flag to the install command which hides the progress animation from stdout.

Issue: #14

### DIFF
--- a/Sources/Swiftly/Install.swift
+++ b/Sources/Swiftly/Install.swift
@@ -59,6 +59,9 @@ struct Install: SwiftlyCommand {
         help: "Verify (or not) the toolchain's PGP signature before proceeding with installation."
     )
     var verify = true
+    
+    @Flag(name: .shortAndLong, help: "Hides the install progress animation.")
+    var quiet = false
 
     @Option(
         help: ArgumentHelp(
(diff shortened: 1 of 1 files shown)


cmcgee1024:
Thanks for your contribution.

There are other progress animations, such as self-update. I wonder if this should be abstracted out to one of the `GlobalOptions` in `Swiftly.swift`. But then, quiet is the opposite of verbose, which is already there so there will need to be some consideration about the interaction of the two. Perhaps they can be unified into a single flag and enum.

In terms of the progress animation. It already appears to be able to detect whether it is running in a terminal context or not, and falls back to a line-oriented output. Are you wanting it to display no progress at all when this option is set? Maybe the detection isn't very effective and leaks escape codes in certain logs? Any other things that should be quieted in the various swiftly commands when the flag is set?

asmeili:
Hi, thanks for taking a look.

During implementation I also considered making this a global flag, however I'd love to hear your input on how this should behave for commands that do not make use of this flag. Is there a way to opt-in this flag only for a particular set of commands without duplicating the flag?

I agree with the verbose flag. Maybe a simple `log` flag that can either be set to `verbose` or `quiet`?

From my understanding of `DynamicProgressAnimation` from `TSCBasic` it looks like, for a terminal output stream, it'll use the animation style. It will then display the progress by sending the output multiple times, which the issuer creator does not seem to want. Instead it looks like they wanted a way to silence this progress output entirely and only have an output of either success or error.

PR #480, opened 2025-12-11
Ensure the macOS SDK exists at install time

justice-adams-apple:
* Ensure a macOS SDK is installed when verifying system prerequisites, if it is not installed. Warn the user at install time

The warning that the end-user will receive is the following (currently):
> WARNING: Could not find a macOS SDK on the system. A macOS SDK is required for the toolchain to work correctly. Please install one via Xcode (https://developer.apple.com/xcode) or run the following command on your machine to install the command line developer tools: `xcode-select --install`.

>If developer tools are located at a non-default location on disk, use `sudo xcode-select --switch path/to/Xcode.app` to     specify the Xcode that you wish to use for command line developer tools.


shahmishal:
@justice-adams-apple can you cherry-pick this on release/1.1?

PR #491, opened 2026-01-13
Set TOOLCHAINS environment variable for macOS when proxying

cmcgee1024:
When using swiftly for macOS, certain tools in the toolchains like swiftPM and Swift Build
internally make use of xcrun, which might pick up a different toolchain that what is currently
selected from swiftly.

Update the TOOLCHAINS environment variable to reflect the version of the Swift toolchain
that is currently selected in swiftly so that the xcrun will pick the same toolchain.

Handle custom SWIFTLY_TOOLCHAINS_DIR by assembling a custom minimal DEVELOPER_DIR
that will cause xcrun to consult the PATH with the toolchain in the beginning and the CLT at
the end to find the necessary tools.

### DIFF
--- a/Sources/MacOSPlatform/MacOS.swift
+++ b/Sources/MacOSPlatform/MacOS.swift
@@ -260,7 +260,7 @@ public struct MacOS: Platform {
         return self.swiftlyToolchainsDir(ctx) / "\(toolchain.identifier).xctoolchain"
     }
 
-    public func updateEnvironmentWithToolchain(_ ctx: SwiftlyCoreContext, _ environment: Environment, _ toolchain: ToolchainVersion) async throws -> Environment {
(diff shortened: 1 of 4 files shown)


xtremekforever:
Thank goodness! I have been waiting for this and manually updating TOOLCHAINS this whole time while using swiftly...

Sources/MacOSPlatform/MacOS.swift
```diff
@@ -185,6 +187,22 @@ public struct MacOS: Platform {
 
         let toolchainDir = self.swiftlyToolchainsDir(ctx) / "\(toolchain.identifier).xctoolchain"
 
+        let bundleID = try await findToolchainBundleID(ctx, toolchain)
+
+        try await fs.remove(atPath: toolchainDir)
```
justice-adams-apple: Will this remove toolchains which existed before swiftly installation?

PR #492, opened 2026-01-16
Prepend SWIFTLY_BIN_DIR to the PATH

### LINKED ISSUE #484: Swiftly's `PATH` updating logic breaks in VS Code
# Description
This issue primarily manifests in VS Code, but will also occur in the terminal if you try to launch a login shell when `$SWIFTLY_BIN_DIR` is already present in `PATH`. I believe this is a macOS only issue because it has a `/usr/bin/swift` script available in `PATH`. However, it could theoretically happen on Linux as well if the user has a `swift` binary somewhere else in `PATH`.

# Steps to Reproduce
1. Launch your terminal which already sources `$SWIFTLY_HOME_DIR/env.sh`
2. Execute `which swift` and see that it reports `$SWIFTLY_BIN_DIR/swift`
3. Launch a login shell within that same terminal window (`zsh -l`)
4. Execute `which swift`

## Actual Result
The second `which swift` will report `/usr/bin/swift` because `/usr/bin` is added to the `PATH` by the login shell. However, for some reason `$SWIFTLY_BIN_DIR` does not get updated and will appear after `/usr/bin`.

## Expected Result
I would expect the second `which swift` to also report `$SWIFTLY_BIN_DIR/swift`.

# Additional Context
The issue appears to lie within the `env.sh` script which checks to see whether or not `$SWIFTLY_BIN_DIR` is already in the `PATH` and, if it is, does nothing. As a user I would expect that it should really be moving `$SWIFTLY_BIN_DIR` to the beginning of `PATH` if I explicitly call the `env.sh` script.

I've worked around this by sourcing `$SWIFTLY_HOME_DIR/env.sh` in my `.zshrc` instead of `.zprofile`. However, the issue will be present for anyone who runs `swiftly init` and lets it update their profile for them.

cmcgee1024:
Instead of checking for the presence of the bin directory in the PATH swiftly should prepend it. This will help to ensure that the proxies aren't shadowed by other tools earlier in the PATH.

Closes #484

### DIFF
--- a/Sources/Swiftly/Init.swift
+++ b/Sources/Swiftly/Init.swift
@@ -113,7 +113,9 @@ struct Init: SwiftlyCommand {
             set -x SWIFTLY_TOOLCHAINS_DIR "\(Swiftly.currentPlatform.swiftlyToolchainsDir(ctx))"
 
             # Remove SWIFTLY_BIN_DIR from PATH if present, then prepend it
-            set -e PATH[(contains -i "$SWIFTLY_BIN_DIR" $PATH)]
+            while set -l index (contains -i "$SWIFTLY_BIN_DIR" $PATH)
+                set -e PATH[$index]
+            end
             set -x PATH "$SWIFTLY_BIN_DIR" $PATH
 
             """


sliemeobn:
🎉

Issue #396, opened 2025-06-30
Improve discoverability of missing required dependencies


Cyberbeni:
They are printed after the init command finishes. -- This can be easily lost if this runs as part of a larger script that installs more stuff.
They are printed after update runs but if we are already using the latest version then they are not printed.

We could have a new subcommand.
We could add automatic checking to env.sh, so people are notified about issues when opening a new terminal. (Sidenote: the installation guide could be more clear that sourcing env.sh should be added to zshrc/bashrc/etc.,)

PR #496, opened 2026-01-24
Add --post-install-file option to init subcommand too

Cyberbeni:
partial fix for #396 

Someone will also have to update the website to suggest using this option after a version has been released: https://www.swift.org/install/linux/

I couldn't get it to build due to `error: 'archive.h' file not found` but the new code shouldn't have any issues.

### DIFF
--- a/Sources/Swiftly/Init.swift
+++ b/Sources/Swiftly/Init.swift
@@ -45,11 +45,20 @@ struct Init: SwiftlyCommand {
     var skipInstall: Bool = false
     @Flag(help: "Quiet shell follow up commands")
     var quietShellFollowup: Bool = false
+    @Option(
+        help: ArgumentHelp(
+            "A file path to a location for a post installation script",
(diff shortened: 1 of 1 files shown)


cmcgee1024:
Thank you for raising this issue and tackling the problem. I've been considering a different approach.

The outputted post install command from `swiftly init` and `swiftly install` can get lost in a terminal buffer. The post install file can similarly be lost or forgotten.

What would be better is something that's more durable, where the user can request it on-demand at any time. If there was a subcommand for swiftly that's responsible for checking system dependency mismatches, and optionally running a command (likely under sudo) to bring in those dependencies could be a better user experience. The `swiftly init` workflow could reuse that functionality too, and if it were to fail for any reason, then the user is directed to this system dependency management subcommand.

Some consideration will be needed in allowing only certain types of commands to be run as root. Perhaps an allow-list or regular expression could help there.

PR #500, opened 2026-01-30
Prevent wrapping of post installation script in swiftly init

cmcgee1024:


Sources/Swiftly/Init.swift
```diff
@@ -304,7 +304,10 @@ struct Init: SwiftlyCommand {
         }
 
         if let postInstall {
-            await ctx.message(Messages.postInstall(postInstall))
+            // This is a plain print instead of a ctx.message to avoid line wrapping of the
+            // post install script. When it is wrapped then it is harder to copy and paste
+            // the contents from the terminal.
+            print("\(Messages.postInstall(postInstall))")
```
justice-adams-apple: would it make more sense to just pretend a newline to this message and still rely on ctx.message? Just thinking about if a user is using json output

justice-adams-apple (approved):
LGTM

cmcgee1024:
I generally like the idea of escaping the newlines of commands so that each argument can be considered. In this case it's probably not as important to most users each individual package as it is that there are packages that need to be installed by the system package manager, so it's probably not as important to devote a line to each package.

PR #284, opened 2025-03-31
Add support for Nushell env var settings

MaxDesiatov:
Currently, swiftly incorrectly generates Bash env var scripts when it is installed in Nushell.

PR #507, opened 2026-02-19
Add support for Nushell and Murex

jobukkit:
This patch adds support for automatically configuring Nushell and Murex shell profiles in addition to the current support for Bash/Zsh/Fish (and makes the README list all these under "Platform support").

Also, during `init` the user is now told which shell profile file is getting added to, and during `self-uninstall` Fish users are no longer left with an empty `swiftly.fish` file (it's deleted if the user didn't add anything to it).

I have manually tested initializing, using, and uninstalling this version of swiftly under both new shells on both MacOS and Linux. I haven't ran the automated tests. Nushell and Murex binaries would have to be added to the test runner.

Obsoletes https://github.com/swiftlang/swiftly/pull/284.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ For more detailed usage guides there is [documentation](https://swiftpackageinde
 
 ## Platform support
 
-swiftly is supported on Linux and macOS. For more detailed information about swiftly's intended features and implementation, check out the [design document](DESIGN.md).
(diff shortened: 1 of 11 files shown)


cmcgee1024 (changes requested):

jobukkit:
Following the release of 1.1.4, I have updated the patch for the current version of the repository. Please let me know if there is still something I need to do.

Issue #490, opened 2026-01-08
Swiftly install stuck while fetching on Linux


WedgeSparda:
Hi, I'm trying to set Swift in my Linux machine (Bazzite, a flavour of Fedora). I followed the [instructions](https://www.swift.org/install/linux/) and set my shell to use `swiftly` command, but when I run `swiftly install latest` the shell gets stuck here:

```
Swiftly is currently unlinked and will not manage the active toolchain. You can run
the following command to link swiftly to the active toolchain:

    $ swiftly link


Fetching the latest stable Swift release...
```

I have to cancel the process if I want to run `swiftly link`, and when I do that, then I get this error:

```
Error: Swiftly couldn't determine the toolchain version to install. 
Please set a version like this and try again: `swiftly install latest`
```

I'm not sure if I'm missing something from the installation process.

WedgeSparda:
Ah, found the reason of the problem. 

Swiftly was stuck in command `yum list installed binutils`, so I run that command in other terminal and saw that a prompt was being raising to confirm a step. I confirmed everything, run `swiftly install latest` again and it worked.

PR #520, opened 2026-04-09
fix: swiftly is stuck during dependency check on Fedora

terraputix:
Fix for https://github.com/swiftlang/swiftly/issues/490.

The underlying issue was that using `.discarded` instead of `.string(limit: 100 * 1024)` will hang when waiting for user input. I verified this by `run`ning a bash script which waits for user input instead of `yum list installed` (from the changed line). 
Since `result.standardOutput` is not currently being used the `limit: 100 * 1024` could be smaller, but for simplicity I used the same value as in the `apt-get` version.

Moreover, `yum list installed` is not fully correct. Compare the following:

```bash
> yum list installed libuuid-devel
Updating and loading repositories:
Repositories loaded.
Available packages
libuuid-devel.i686   2.40.4-10.fc42 updates
libuuid-devel.x86_64 2.40.4-10.fc42 updates
> echo $?
0
> yum list --installed libuuid-devel
No matching packages to list
> echo $?
1
```

This shows that using `--installed` instead of `installed` will correctly return a non 0 exit code if a package is not installed.

### DIFF
--- a/Sources/LinuxPlatform/Linux.swift
+++ b/Sources/LinuxPlatform/Linux.swift
@@ -330,7 +330,7 @@ public struct Linux: Platform {
                 }
                 return false
             case "yum":
-                let result = try await run(.name("yum"), arguments: ["list", "installed", package], output: .discarded)
+                let result = try await run(.name("yum"), arguments: ["list", "--installed", package], output: .string(limit: 100 * 1024))
                 return result.terminationStatus.isSuccess
             default:
                 return true


Sources/LinuxPlatform/Linux.swift
```diff
@@ -330,7 +330,7 @@ public struct Linux: Platform {
                 }
                 return false
             case "yum":
-                let result = try await run(.name("yum"), arguments: ["list", "installed", package], output: .discarded)
+                let result = try await run(.name("yum"), arguments: ["list", "--installed", package], output: .string(limit: 100 * 1024))
```
cmcgee1024, line 333: Unfortunately, this may not be working with Amazon Linux 2, which doesn't have dnf options like `--installed`.

terraputix: What would your preferred solution for this look like? I guess it would be simplest to check the output of the command for a correctly installed package (similar to what is done for the apt-get version). 
Alternatively, `dnf` could be called with `yum` as a fallback?

PR #531, opened 2026-04-30
Print errors to stderr

keith:
Previously these cases went to stdout, so error messages like:

```
.../.swift-version`
uses toolchain version 6.1, but it doesn't match any of the installed
toolchains. You can install the toolchain with `swiftly install`.
```

Would be lost unless you included captured stdout as well


ktoso:
Logging errors to stderr makes sense to me, any reason to not merge this @cmcgee1024 ?

PR #540, opened 2026-05-15
Use: improve error message when .swift-version toolchain is missing

cmcgee1024:
When a swift version file references an uninstalled toolchain, the error
now suggests a compatible installed version (same major.minor) as a
concrete `swiftly use` command, rather than a generic placeholder.
Also clarifies the message to distinguish "not installed" from "doesn't match".


etcwilde (approved):

etcwilde (approved):
Should probably have a test, but LGTM.

PR #561, opened 2026-06-13
Handle install 6.4-snapshot properly

### LINKED ISSUE #560: swiftly install 6.4-snapshot fails because url schema change
This is expected to work:

```
ktoso@cherudim [SIGINT] [11:43:48] [~/code/swiftly] [wip-handle-6.4 *]
-> % swiftly install 6.4-snapshot
Fetching the latest 6.4 development branch snapshot...
-----------------------------
A new release of swiftly is available.
Please run `swiftly self-update` to update.
-----------------------------
Error: Unexpected response, expected status code: ok, response: undocumented(statusCode: 404, OpenAPIRuntime.UndocumentedPayload(headerFields: [(field: Server: Apple, next: 65535), (field: Date: Sat, 13 Jun 2026 02:43:54 GMT, next: 65535), (field: Content-Type: text/html, next: 65535), (field: Content-Length: 146, next: 65535), (field: Strict-Transport-Security: max-age=31536000; includeSubdomains, next: 65535), (field: X-Frame-Options: SAMEORIGIN, next: 65535), (field: X-XSS-Protection: 1; mode=block, next: 65535), (field: Age: 2, next: 65535), (field: Via: https/1.1 jptyo5-edge-mx-010.ts.apple.com (acdn/316.16515), https/1.1 jptyo5-edge-fx-028.ts.apple.com (acdn/316.16515), next: 65535), (field: X-Cache: hit-stale, miss, next: 65535), (field: CDNUUID: f0997761-ef15-4f9d-ac2e-03889f7307d5-156837625, next: 65535), (field: Connection: keep-alive, next: 65535)], body: Optional(OpenAPIRuntime.HTTPBody)))
```

but doesn't because the url changed to `6.4.x`

ktoso:
swift.org changed the url scheme and 6.4 now needs to become 6.4.x.

We should not require users to know about this and installing 6.4 or 6.4-snapshot is totally expected to work.

This fixes the known urls and also adds a fallback for the future;

Options to consider:
- we could remove the fallback... I'm just being extra careful about future i guess, perhaps too much?
- we could default all versions from 6.4 onwards to add the .x always, but that's pretty spectulative so i didn't do that.

Current scheme adds the .x for 6.4 and for others it would attempt a fallback to .x but now try it first.

Resolves https://github.com/swiftlang/swiftly/issues/560

---

before:

```
-> % swiftly install 6.4-snapshot
Fetching the latest 6.4 development branch snapshot...
-----------------------------
A new release of swiftly is available.
Please run `swiftly self-update` to update.
-----------------------------
Error: Unexpected response, expected status code: ok, response: undocumented(statusCode: 404, OpenAPIRuntime.UndocumentedPayload(headerFields: [(field: Server: Apple, next: 65535), (field: Date: Sat, 13 Jun 2026 02:43:54 GMT, next: 65535), (field: Content-Type: text/html, next: 65535), (field: Content-Length: 146, next: 65535), (field: Strict-Transport-Security: max-age=31536000; includeSubdomains, next: 65535), (field: X-Frame-Options: SAMEORIGIN, next: 65535), (field: X-XSS-Protection: 1; mode=block, next: 65535), (field: Age: 2, next: 65535), (field: Via: https/1.1 jptyo5-edge-mx-010.ts.apple.com (acdn/316.16515), https/1.1 jptyo5-edge-fx-028.ts.apple.com (acdn/316.16515), next: 65535), (field: X-Cache: hit-stale, miss, next: 65535), (field: CDNUUID: f0997761-ef15-4f9d-ac2e-03889f7307d5-156837625, next: 65535), (field: Connection: keep-alive, next: 65535)], body: Optional(OpenAPIRuntime.HTTPBody)))
```

after

```
-> % ./.build/debug/swiftly install 6.4-snapshot
Fetching the latest 6.4 development branch snapshot...
Installing 6.4.x-snapshot-2026-06-10
                                                                                                        Downloading 6.4.x-snapshot-2026-06-10
 2% [====------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------]
Downloaded 42.1 MiB of 1855.3 MiB^C
```

### DIFF
--- a/Sources/SwiftlyCore/HTTPClient.swift
+++ b/Sources/SwiftlyCore/HTTPClient.swift
@@ -445,7 +445,7 @@ extension SwiftlyWebsiteAPI.Components.Schemas.DevToolchainForArch {
                     message: "malformatted release branch: \"\(majorString).\(minorString)\"")
             }
             let patch = match.output.3.map(String.init)
-            branch = .release(major: major, minor: minor, patch: patch)
+            branch = .releaseNormalized(major: major, minor: minor, patch: patch)
         } else {
             branch = .main
         }
@@ -464,19 +464,6 @@ public struct DownloadNotFoundError: LocalizedError {
 
     public init(url: URL) {
         self.url = url
-    }
-}
-
-/// Returns true for branches whose `swift.org` snapshot path uses the `X.Y.x`
-/// shape rather than `X.Y`. As of writing, the 6.4 development branch is
-/// published at `/install/dev/6.4.x/...` while older release branches (6.0–6.3)
-/// remain at `/install/dev/X.Y/...`.
-private func sourceBranchNeedsXPatchSuffix(_ branch: ToolchainVersion.Snapshot.Branch) -> Bool {
-    switch branch {
-    case .release(major: 6, minor: 4, patch: nil):
-        return true
-    default:
-        return false
     }
 }
 
@@ -591,34 +578,13 @@ public struct SwiftlyHTTPClient: Sendable {
         {
         case .main:
             .init(.main)
-        case let .release(major, minor, nil) where sourceBranchNeedsXPatchSuffix(branch):
-            .init("\(major).\(minor).x")
         case let .release(major, minor, patch):
             .init("\(major).\(minor)\(patch.map { ".\($0)" } ?? "")")
         }
 
-        let devToolchains: SwiftlyWebsiteAPI.Components.Schemas.DevToolchains
-        do {
-            devToolchains = try await self.httpRequestExecutor.getSnapshotToolchains(branch: sourceBranch, platform: platformId)
-        } catch let originalError {
-            // swift.org changed url schemes and 6.4+ versions need `.x` patch suffix.
-            // For known versions which need this we request the right path based on `sourceBranch`,
-            // but just in case, if an initial lookup fails re-try with `.x` suffix if it applies.
-            guard case let .release(major, minor, nil) = branch else {
-                // We can't append an .x patch if a patch version was already specified.
-                throw originalError
-            }
-
-            // Last resort, retry with .x patch suffix
-            guard let devToolchainsPatched =
-                  try? await self.httpRequestExecutor.getSnapshotToolchains(
-                    branch: .init("\(major).\(minor).x"), platform: platformId) else {
-                // Ignore the last-resort failure, report the original failure
-                throw originalError
-            }
-
-            devToolchains = devToolchainsPatched
-        }
+        let devToolchains = try await self.httpRequestExecutor.getSnapshotToolchains(
+            branch: sourceBranch, platform: platformId
+        )
 
         let arch = a ?? cpuArch.value2
 
--- a/Sources/SwiftlyCore/HTTPClient.swift
+++ b/Sources/SwiftlyCore/HTTPClient.swift
@@ -467,6 +467,19 @@ public struct DownloadNotFoundError: LocalizedError {
     }
 }
 
+/// Returns true for branches whose `swift.org` snapshot path uses the `X.Y.x`
+/// shape rather than `X.Y`. As of writing, the 6.4 development branch is
+/// published at `/install/dev/6.4.x/...` while older release branches (6.0–6.3)
+/// remain at `/install/dev/X.Y/...`.
+private func sourceBranchNeedsXPatchSuffix(_ branch: ToolchainVersion.Snapshot.Branch) -> Bool {
+    switch branch {
+    case .release(major: 6, minor: 4, patch: nil):
+        return true
+    default:
+        return false
+    }
+}
+
 /// HTTPClient wrapper used for interfacing with various REST APIs and downloading things.
 public struct SwiftlyHTTPClient: Sendable {
     public let httpRequestExecutor: HTTPRequestExecutor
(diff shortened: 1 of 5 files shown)


ktoso (commented):
I’ll do some more testing if this doesn’t break anything after the weekend;

Sources/SwiftlyCore/HTTPClient.swift
```diff
@@ -467,6 +467,19 @@ public struct DownloadNotFoundError: LocalizedError {
...
+/// published at `/install/dev/6.4.x/...` while older release branches (6.0–6.3)
+/// remain at `/install/dev/X.Y/...`.
+private func sourceBranchNeedsXPatchSuffix(_ branch: ToolchainVersion.Snapshot.Branch) -> Bool {
+    switch branch {
+    case .release(major: 6, minor: 4, patch: nil):
+        return true
+    default:
+        return false
+    }
+}
```
ktoso: So here's where we could take a policy decision to just assume future versions follow this .x style or be conservative like it is now and just special case the 6.4... WDYT is better? I guess we'll find out in the future either way heh


Maybe flip it for v >= 6.4 -> true?

etcwilde: So to answer your question, it is expected that the `.x` will be part of any release development branch, aligning with the branch name on the Swift repository on GitHub, so this can probably be reduced to `v >= 6.4`.

That said, I wonder if it would make sense to do this in the `ToolchainVersion`? If it's a snapshot, and the version is 6.4 or newer, the patch version automatically chooses `x`. Then we don't have to add the additional `.release(major, minor, nil)` below, and I think we can drop the logic in the catch portion of the `getSnapshotToolchains` below. The `patch` string will just return `x`.

etcwilde (changes requested):
I talked with @shahmishal about this earlier today. The `6.4.x` is intentional, specifically so that folks don't expect  `6.4.0` when they pull down a `6.4` snapshot from swift.org.
Since Swiftly already gives you the latest, the 6.4.x behavior is already expected.

Since the website is going to start calling the nightlies with `.x`, but swiftly already makes it apparent that a snapshot is a snapshot in the version selector, I think that it's valuable to accept both.

I think it might be interesting to automatically have the `snapshot` set the patch to "x" if it's 6.4 or greater and one wasn't specified explicitly. Then we should update the `parseReleaseSnapshotWithPatch` test to use 6.4 for each, and include a `6.4-snapshot` case that should also map to a `.release(major: 6, minor: 4, patch: "x")` snapshot version to ensure that it's handling the mapping correctly.

ktoso:
Gotcha, let me rework this a bit then. Bottom line is I think we do want to accept the `6.4-snapshot` if the `.x` was not specified, but we won't do the `nil` case in the enum but model it with defaulting to the `x` instead if I understood you correctly. I'll take a stab at it

ktoso:
This ended up much cleaner, please have another look @etcwilde -- thanks!

Tests/SwiftlyTests/ToolchainSelectorTests.swift
```diff
@@ -63,6 +63,20 @@ import Testing
         try runTest(.snapshot(branch: .release(major: 5, minor: 7, patch: "x"), date: nil), parses)
     }
 
+    /// From Swift 6.4 onward the swift.org URL scheme uses `X.Y.x`, so a bare
+    /// `6.4-snapshot` (no explicit patch) is normalized to `patch: "x"`.
+    @Test func parseReleaseSnapshotWithoutPatch() throws {
```
etcwilde: Okay, I think we need to verify that `5.7.x-snapshot` and `5.7-snapshot` don't evaluate to the same thing and that `6.4-snapshot` and `6.4.x-snapshot` do.

Since toolchains moving forward will all carry the `.x` patch, I'm inclined to say that we should take the `parseReleaseSnapshot` name for the new test verifying the behaviour of `6.4-snapshot` and `6.4.x-snapshot`, and rename the old test function something liken `parseDeprecatedReleaseSnapshot`.

Then we merge this with the `parseReleaseSnapshotWithPatch` test above.

```swift
@Test func parseDeprecatedReleaseSnapshot() throws {
    var parses = [
        "5.7-snapshot",
        "5.7-SNAPSHOT",
        "5.7-DEVELOPMENT-SNAPSHOT",
        "swift-5.7-snapshot",
        "swift-5.7-SNAPSHOT",
        "swift-5.7-DEVELOPMENT-SNAPSHOT",
    ]
    try runTest(.snapshot(branch: .release(major: 5, minor: 7, patch: nil), date: nil), parses)

    parses = [
        "5.7.x-snapshot",
        "5.7.x-SNAPSHOT",
        "5.7.x-DEVELOPMENT-SNAPSHOT",
        "swift-5.7.x-snapshot",
        "swift-5.7.x-SNAPSHOT",
        "swift-5.7.x-DEVELOPMENT-SNAPSHOT",
    ]
    try runTest(.snapshot(branch: .release(major: 5, minor: 7, patch: "x"), date: nil), parses)
}

@Test func parseReleaseSnapshot() throws {
    let parses = [
        "6.4-snapshot",
        "6.4.x-snapshot",
        "6.4-SNAPSHOT",
        "6.4.x-SNAPSHOT",
        "6.4-DEVELOPMENT-SNAPSHOT",
        "6.4.x-DEVELOPMENT-SNAPSHOT",
        "swift-6.4-snapshot",
        "swift-6.4.x-snapshot",
        "swift-6.4-SNAPSHOT",
        "swift-6.4.x-SNAPSHOT",
        "swift-6.4-DEVELOPMENT-SNAPSHOT",
        "swift-6.4.x-DEVELOPMENT-SNAPSHOT",
    ]
    try runTest(.snapshot(branch: .release(major: 6, minor: 4, patch: "x"), date: nil), parses)
}
```

This should cover everything in `parseReleaseSnapshot`, `parseReleaseSnapshotWithPatch` and `parseReleaseSnapshotWithoutPatch`.
Thoughts?

etcwilde (commented):
I like that a lot better. Thanks for making the changes.

I'd like to add verification that `6.4.x-snapshot` and `6.4-snapshot` evaluate to the same thing, and that `5.7.x-snapshot` and `5.7-snapshot` don't.

Tests/SwiftlyTests/ToolchainSelectorTests.swift
ktoso: Sure, sounds good -- and I like the having the "deprecated" for the old one, I'll follow up soon

ktoso:
I think I captured the changes you requested @etcwilde, please have a look

ches:
It seems like stable releases need similar treatment too… **edit**: looks like #594 is addressing this.

```
$ swiftly update latest
Update Swift 6.2.4 -> Swift 6.4.0?
Proceed? (Y/n): 

Y
Installing Swift 6.4.0
Error: Swift 6.4.0 does not exist at URL https://download.swift.org/swift-6.4-release/xcode/swift-6.4-RELEASE/swift-6.4-RELEASE-osx.pkg, exiting

$ swiftly --version
1.1.3
```

The scheme is:

<https://download.swift.org/swift-6.4.0-release/xcode/swift-6.4.0-RELEASE/swift-6.4.0-RELEASE-osx.pkg>

shahmishal:
Swiftly 1.1.4 was released to address @ches your issue.

PR #565, opened 2026-06-20
Add FreeBSD platform support

networkextension:
## Summary

Introduces a \`FreeBSDPlatform\` target so swiftly compiles, runs, and passes its full test suite on FreeBSD, alongside the existing Linux and macOS platforms.

## Approach

- New **\`FreeBSDPlatform\`** (\`struct FreeBSD: Platform\`), modeled on \`LinuxPlatform\` since FreeBSD's Swift toolchains are tarball-based like Linux. \`detectPlatform\` returns a single \`PlatformDefinition.freebsd\`; the install/extract/directory logic reuses the portable POSIX/Foundation paths.
- \`Swiftly\` and \`TestSwiftly\` select it via \`#elseif os(FreeBSD)\`, and the POSIX \`os(macOS) || os(Linux)\` guards in \`SwiftlyCore\` (\`ProcessInfo\`, \`Platform\`, \`Platform+Process\`, \`Terminal\`) now include FreeBSD.
- Adds \`PlatformDefinition.freebsd\`.
- \`FreeBSDPlatform\` is gated with \`.when(platforms: [.custom("freebsd")])\` (thanks @kkebo), matching the \`LinuxPlatform\`/\`MacOSPlatform\` entries — \`PackageDescription\` has no \`.freebsd\` case, but \`.custom("freebsd")\` works, so it stays out of the graph on non-FreeBSD builds.
- The release-packaging tool's \`run()\` throws on FreeBSD for now (the Linux release path is musl-SDK specific).
- \`verifySystemPrerequisitesForInstall\`: uses \`pkg info -e\` and FreeBSD dependency names (\`libuuid\`, \`python311\`, \`sqlite3\`) from swift-installer-scripts.
- \`getShell\`: falls back to \`/bin/sh\` (not \`/bin/bash\`, which is not the default shell on FreeBSD).

## Testing

All **170 SwiftlyTests pass** on two FreeBSD boards:
- **FreeBSD 15.1 / aarch64** (NXP DPAA2, 16-core, native Swift 6.3.2-RELEASE toolchain)
- **FreeBSD 14.4 / aarch64** (NXP DPAA2, native Swift 6.3.2-RELEASE toolchain)

CI runs via Cloudflare Tunnel to self-hosted boards (see workflow). HTTPClientTests (21 tests) require real network access to swift.org and also pass.

Tests use \`Subprocess.run\` instead of \`Foundation.Process\` on FreeBSD — \`Process.run()\` triggers a \`_CFRunLoopCreate\` assertion when called from a libdispatch worker thread on FreeBSD, causing SIGTRAP. \`Subprocess\` uses \`posix_spawn\` internally and is concurrency-safe.

> Note: swift.org doesn't yet publish FreeBSD toolchains, so this is foundational platform support — it makes swiftly itself build/run on FreeBSD and will enable toolchain management once FreeBSD builds are published.


kkebo:
> `.when(platforms: [.freebsd])` can't be used

You can use `.when(platforms: [.custom("freebsd")])` instead.

networkextension:
Thanks @kkebo — that's much cleaner. Applied in 4f25b50: the three `FreeBSDPlatform` dependencies now use `.when(platforms: [.custom("freebsd")])`, matching the `LinuxPlatform`/`MacOSPlatform` entries, so it's no longer pulled into the graph on non-FreeBSD builds. I'll update the PR description note accordingly.

Sources/FreeBSDPlatform/FreeBSD.swift
```diff
@@ -0,0 +1,527 @@
...
+
+        guard !missingPackages.isEmpty else {
+            return nil
+        }
+
+        return "\(manager) -y install \(missingPackages.joined(separator: " "))"
+    }
+
+    public func isSystemPackageInstalled(_ manager: String?, _ package: String) async -> Bool {
+        do {
```
jakepetroules, line 112: This should be implemented in terms of `pkg` rather than Linux package managers.

Sources/FreeBSDPlatform/FreeBSD.swift
```diff
@@ -0,0 +1,527 @@
...
+            """
+
+            throw SwiftlyError(message: msg)
+        }
+    }
+
+    public func verifySystemPrerequisitesForInstall(
+        _ ctx: SwiftlyCoreContext, platformName: String, version _: ToolchainVersion,
+        requireSignatureValidation: Bool
+    ) async throws -> String? {
```
jakepetroules, line 70: All of these Linux packages should be replaced with what's needed for FreeBSD.

You can find a list to start with on https://github.com/swiftlang/swift-installer-scripts/blob/main/platforms/FreeBSD/makePackage#L108

Sources/FreeBSDPlatform/FreeBSD.swift
```diff
@@ -0,0 +1,527 @@
...
+    }
+
+    public func getShell() async throws -> String {
+        let userName = ProcessInfo.processInfo.userName
+        if let entry = try await sys.getent(database: "passwd", key: userName).entries().first {
+            if let shell = entry.last { return shell }
+        }
+
+        // Fall back on bash
+        return "/bin/bash"
```
jakepetroules: I think bash isn't installed on FreeBSD by default, right? maybe the default should be /bin/sh

jakepetroules (changes requested):

jakepetroules:
Awesome to see more FreeBSD progress!

networkextension:
@jakepetroules Thanks! Here is a full status update.

## Test results on FreeBSD 15.1 aarch64 (Swift 6.3.2-RELEASE, NXP DPAA2 board)

### HTTPClientTests — all pass with NIOSSL CA fix

The prior failure (`CERTIFICATE_VERIFY_FAILED`) was a missing FreeBSD branch in `swift-nio-ssl`'s `platformDefaultConfiguration()` — zero CA trust roots were loaded. With a two-line fix, all 6 tests pass over real HTTPS to `swift.org`/`download.swift.org`:

```
✔ Test getSwiftOrgGPGKeys() passed after 0.700 seconds.
✔ Test getSwiftToolchain() passed after 109.816 seconds.
✔ Test getSwiftlyLinuxReleases(url:) with 6 test cases passed after 31.636 seconds.
✔ Test getSwiftlyMacOSReleases(url:) with 3 test cases passed after 6.700 seconds.
✔ Test getSwiftlyReleaseMetadataFromSwiftOrg() passed after 0.576 seconds.
✔ Test getToolchainMetadataFromSwiftOrg(_:) with 9 test cases passed after 34.348 seconds.
✔ Suite HTTPClientTests passed after 183.785 seconds.
✔ Test run with 6 tests in 1 suite passed after 183.786 seconds.
```

The NIOSSL fix (`SSLContext.swift` + `LinuxCABundle.swift`):

```diff
-#if os(Linux)
+#if os(Linux) || os(FreeBSD)
```

Plus FreeBSD CA paths: `/usr/local/etc/ssl/cert.pem`, `/etc/ssl/cert.pem`, `/usr/local/share/certs/ca-root-nss.crt`. Will open a PR to `swift-nio-ssl`.

### UpdateTests — all pass

```
✔ Suite UpdateTests passed after 11.360 seconds.
✔ Test run with 13 tests in 2 suites passed after 20.007 seconds.
```

All other suites (PlatformTests, InstallTests, UninstallTests, LinkTests, ListTests, UseTests, etc.) confirmed passing in earlier runs on the same board.

## Dependency patches to build and test on FreeBSD

`swift test` pulls in `swift-nio` and `swift-nio-ssl`, which each need FreeBSD fixes. All are upstream PR candidates.

**swift-nio / `_NIOFileSystem`:**
- `O_TMPFILE` is Linux-only — exclude FreeBSD from `FileDescriptor.OpenOptions.temporaryFile`; keep `currentWorkingDirectory` (`AT_FDCWD` is on FreeBSD)
- xattr (`flistxattr`, `fgetxattr`, `fsetxattr`, `fremovexattr`) — FreeBSD uses `extattr_*`; add `#elseif os(FreeBSD)` stubs returning -1
- `sendfile` — FreeBSD reverses file/socket args and returns bytes via out-param; needs its own `#if os(FreeBSD)` block
- `RENAME_NOREPLACE`, `AT_EMPTY_PATH`, `CNIOLinux_renameat2` — add to CNIOLinux.h FreeBSD section (FreeBSD 15 has these natively)
- `Errno.noData` (`ENODATA`) — not on FreeBSD; guard with `#if !os(FreeBSD)`
- `fts_info` is `int` on FreeBSD (not `u_short`); `DT_*` constants need `|| os(FreeBSD)` on Int32 branch

**swift-nio / NIOCore + NIOPosix:**
- Use `os(OpenBSD) || os(FreeBSD)` blocks (not `os(Linux) || os(FreeBSD)`) — Glibc on FreeBSD doesn't expose the same symbols as on Linux
- kqueue: `EVFILT_EXCEPT` stub; `isEarlyEOFDeliveryWorkingOnThisOS = false`
- CMSG shims via `CNIOOpenBSD_CMSG_*`; BSD `sendfile` via `CNIOOpenBSD.sendfile()`; `UIO_MAXIOV=1024`; `SHUT_*` as literals

**swift-nio-ssl:**
- Add `#elseif os(FreeBSD)` import block in `NIOSSLClientHandler`, `IdentityVerification`, `SubjectAlternativeName`
- Route `inet_ntop` through `CNIOOpenBSD_inet_ntop()` on FreeBSD
- CA trust root fix above

## Related: Synchronization.Mutex deadlock

`Synchronization.Mutex` deadlock on FreeBSD (swiftlang/swift#90057) has a fix PR at swiftlang/swift#90143. That was the root cause of `swift-testing`'s SIGTRAP on FreeBSD.

Happy to open separate upstream PRs to `swift-nio` and `swift-nio-ssl` for the dependency patches.

Sources/FreeBSDPlatform/FreeBSD.swift
networkextension: Done in `da8a452` — `isSystemPackageInstalled` now uses `pkg info -e <name>`, and `verifySystemPrerequisitesForInstall` returns `pkg install <missing>`. All `apt-get`/`dnf`/`yum` paths were removed.

Sources/FreeBSDPlatform/FreeBSD.swift
networkextension: Done in `da8a452` — replaced the Linux package list with the FreeBSD runtime deps from the `makePackage` reference you linked: `libuuid`, `python311`, `sqlite3`. Missing ones are surfaced to the user as `pkg install <names>`.

Sources/FreeBSDPlatform/FreeBSD.swift
networkextension: Done in `da8a452` — `getShell()` reads the user's shell from `getent passwd`, then falls back to `/bin/sh` (bash is not installed by default on FreeBSD).

networkextension:
@jakepetroules thanks for the review — addressed all three in `da8a452`:

- FreeBSD runtime deps (`libuuid`, `python311`, `sqlite3`) per the `makePackage` list you linked, replacing the Linux packages.
- Package detection/install via `pkg info -e` / `pkg install` (no more `apt`/`dnf`/`yum`).
- Default shell falls back to `/bin/sh`.

Separately, the underlying `Synchronization.Mutex` FreeBSD deadlock (swiftlang/swift#90057) now has a fix in swiftlang/swift#90143, which I have cross-built and verified deadlock-free under contention (16 threads × 100k) on FreeBSD 14.4 and 15.1 / aarch64. Re-requesting review.

PR #570, opened 2026-06-27
Reorder getting-started tabs to put macOS first

ahmdyasser:
## Motivation

In the _Getting started with swiftly_ documentation, the platform installation tabs are currently ordered with Linux first, so Linux is the default tab readers see first. macOS is Swift's primary development platform, so this PR reorders the "macOS" tab ahead of the "Linux" tab, making macOS the default selected tab in the `@TabNavigator`.

## Changes

- Swapped the order of the `@Tab("macOS")` and `@Tab("Linux")` blocks in `Documentation/SwiftlyDocs.docc/getting-started.md`. No content within either tab was modified.

## Notes

- Documentation-only change; no code or tests affected.
- Followed the Swift contribution guidelines (<https://www.swift.org/contributing/>): concise title with rationale in the body, clean commit, no in-source attribution.

### DIFF
--- a/Documentation/SwiftlyDocs.docc/getting-started.md
+++ b/Documentation/SwiftlyDocs.docc/getting-started.md
@@ -5,21 +5,6 @@ Start using swiftly and Swift.
 To get started with swiftly you can download it from [swift.org](https://swift.org/download), and extract the package.
 
 @TabNavigator {
-    @Tab("Linux") {
-        If you are using Linux then you can download the binary:
-
-        ```
-        curl -L https://download.swift.org/swiftly/linux/swiftly-$(uname -m).tar.gz > swiftly.tar.gz
-        tar zxf swiftly.tar.gz
-        ```
-
-        Now run swiftly init to finish the installation:
-
-        ```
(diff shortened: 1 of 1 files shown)


plemarquand:
I don't know that the claim that "MacOS is Swiftly's primary development platform" holds true. macOS users have typically installed toolchains via Xcode. Swiftly is IMO most useful on Linux and remains the primary way to install toolchains there.

ahmdyasser:
@plemarquand 
The reason I raised the PR because when I wanted to install Swiftly I blindly copied and pasted the install script on my Mac and it -obviously- didn't work. I assumed since it's "Swift" then the default should be Mac. I might be wrong because I'm not aware of the Swift Linux users. If this is more of opinionated thing, feel free to close the PR. Thanks for the reply.

compnerd:
I agree with @plemarquand in the questioning of the veracity of the claim that macOS is the primary platform. This is a user specific bit: windows is my primary platform for swift work.

We could leave the command tabbed and select the tab based on the host os like we do with the download page for swift.

PR #530, opened 2026-04-23
Fedora41: Adding fedora41 support

etcwilde:
Teaching Swiftly to recognize Fedora41 and adding it to the pull-request testing.

https://github.com/swiftlang/swiftly/issues/438

PR #575, opened 2026-07-20
Add support for Ubuntu 26.04 and Debian 13

### LINKED ISSUE #569: Add support for Debian 13
I’m using Debian 13. Although Swiftly doesn’t officially support Debian 13, I’ve configured it to download the Debian 12 binaries as shown below.

```json
{
  "platform" : {
    "nameFull" : "debian12",
    "namePretty" : "Debian GNU\/Linux 12",
    "name" : "debian12"
  },
  ...
  "version" : "1.1.2"
}
```

It works fine in most cases, but `swift repl` does not work. This is because it is dynamically linked to `libpython3.11.so.1.0`, and Python 3.13 is the default on Debian 13.

Therefore, could you please add support for Debian 13?

It appears that the main-snapshot toolchain for Debian 13 has already begun to be published.

- https://github.com/swiftlang/swift-org-website/blob/main/_data/builds/development/debian13.yml
- https://github.com/swiftlang/swift-org-website/blob/main/_data/builds/development/debian13-aarch64.yml

kkebo:
Closes #569

Corresponding PR swiftlang/swift-org-website#1478 is required to make `swiftly install main-snapshot` work and pass the `getToolchainMetadataFromSwiftOrg` test.

I have updated the list of dependent packages, using swiftlang/swift-docker as a reference.

<details><summary>diff nightly-main/ubuntu/{24,26}.04/Dockerfile</summary>

```diff
1c1
< FROM ubuntu:24.04
---
> FROM ubuntu:resolute
7a8
>     binutils-gold \
10d10
<     unzip \
16c16
<     libgcc-13-dev \
---
>     libgcc-15-dev \
19c19
<     libstdc++-13-dev \
---
>     libstdc++-15-dev \
36c36
< ARG OS_MAJOR_VER=24
---
> ARG OS_MAJOR_VER=26
```

</details>

<details><summary>diff nightly-main/debian/{12,13}/Dockerfile</summary>

```diff
1c1
< FROM debian:12
---
> FROM debian:13
20c20
<     libstdc++-12-dev \
---
>     libstdc++-14-dev \
25,26c25,26
< ARG SWIFT_PLATFORM=debian12
< ARG SWIFT_VERSION=6.2
---
> ARG SWIFT_PLATFORM=debian13
> ARG SWIFT_VERSION=6.4
```

</details>

I didn't add both platforms to the `tests-selfhosted` and `tests-bootstrapped` CI jobs yet because there are no released Swiftly builds for them and no `swiftlang/swift:nightly-main-{debian13,resolute}` images on Docker Hub.

Other references:

- https://github.com/swiftlang/swiftly/pull/530

### DIFF
--- a/Sources/LinuxPlatform/Linux.swift
+++ b/Sources/LinuxPlatform/Linux.swift
@@ -11,6 +11,7 @@ typealias fs = SwiftlyCore.FileSystem
 /// TODO: replace dummy implementations
 public struct Linux: Platform {
     let linuxPlatforms: [PlatformDefinition] = [
+        .ubuntu2604,
         .ubuntu2404,
         .ubuntu2204,
         .ubuntu2004,
@@ -20,6 +21,7 @@ public struct Linux: Platform {
         .rhel9,
         .amazonlinux2,
         .debian12,
+        .debian13,
     ]
 
     public init() {}
@@ -164,6 +166,28 @@ public struct Linux: Platform {
                 "tzdata",
                 "zlib1g-dev",
             ]
+        case "ubuntu2604":
+            [
+                "binutils",
+                "binutils-gold",
+                "git",
+                "unzip",
+                "zip",
+                "gnupg2",
+                "libc6-dev",
+                "libcurl4-openssl-dev",
+                "libedit2",
+                "libgcc-15-dev",
+                "libpython3-dev",
+                "libsqlite3-0",
+                "libstdc++-15-dev",
+                "libxml2-dev",
+                "libncurses-dev",
+                "libz3-dev",
+                "pkg-config",
+                "tzdata",
+                "zlib1g-dev",
+            ]
         case "amazonlinux2":
             [
                 "binutils",
@@ -237,6 +261,25 @@ public struct Linux: Platform {
                 "unzip",
                 "zip",
             ]
+        case "debian13":
+            [
+                "binutils", // binutils-gold is a virtual package that points to binutils
+                "libicu-dev",
+                "libcurl4-openssl-dev",
+                "libedit-dev",
+                "libsqlite3-dev",
+                "libncurses-dev",
+                "libpython3-dev",
+                "libxml2-dev",
+                "pkg-config",
+                "uuid-dev",
+                "tzdata",
+                "git",
+                "gcc",
+                "libstdc++-14-dev",
+                "unzip",
+                "zip",
+            ]
         default:
             []
         }
@@ -252,13 +295,15 @@ public struct Linux: Platform {
             "apt-get"
         case "ubuntu2404":
             "apt-get"
+        case "ubuntu2604":
+            "apt-get"
         case "amazonlinux2":
             "yum"
         case "ubi9":
             "dnf"
         case "fedora39", "fedora41":
             "dnf"
-        case "debian12":
+        case "debian12", "debian13":
             "apt-get"
         default:
             nil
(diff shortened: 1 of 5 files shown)


kkebo:
I fixed the `getToolchainsMetadataFromSwiftOrg` test.

Sources/LinuxPlatform/Linux.swift
```diff
@@ -237,6 +261,25 @@ public struct Linux: Platform {
                 "unzip",
                 "zip",
             ]
+        case "debian13":
+            [
+                "binutils", // binutils-gold is a virtual package that points to binutils
```
etcwilde: `binutils-gold` is needed to install the gold linker. With just `binutils`, `/usr/bin/ld.gold` isn't installed.

Tests/SwiftlyTests/HTTPClientTests.swift
```diff
@@ -134,18 +134,26 @@ import Testing
         let platform: PlatformDefinition
         let architectures: [SwiftlyWebsiteAPI.Components.Schemas.Architecture]
         let branches: [ToolchainVersion.Snapshot.Branch]
+        let isReleased: Bool
```
etcwilde: The `HTTPClientTests` is meant to ensure that we are generating the right requests against the Swift.org API. It's not really meant to verify that we can generate every URL for everything under the sun, so if we're generating test cases that we're skipping, that doesn't really make sense. I'm also concerned about setting up the Swiftly tests to fail when something happens outside of the Swiftly repository (like swift.org releasing toolchains).

I see two alternative paths that I would be happy to accept
 1. Generating the URLs is mostly distro-agnostic, so I'm fine if we want to drop distros that don't have a release yet.
 2. Alternatively, we can split `getToolchainMetadataFromSwiftOrg` into two separate tests. One to test the release toolchain URLs and one to test the snapshot toolchain URLs. Then we would only include Debian 13 and Ubuntu 26.04 in the snapshot test and omit it from the release test until there is a release.

etcwilde (changes requested):
Thanks for the contribution. A couple of points that I would like to see changed:
 - We still need to install the gold linker, which doesn't look like it happens with the `binutils` package.
 - I would like to rethink the HTTP tests to either just omit the distros that are purely snapshot URL request tests, or split the release request testing from the snapshot request testing.

Sources/LinuxPlatform/Linux.swift
kkebo: Sorry, I missed that. On Debian 12, `binutils-gold` was a virtual package provided by `binutils`, but it looks like that’s no longer the case in Debian 13.

- https://packages.debian.org/bookworm/binutils-gold
- https://packages.debian.org/trixie/binutils-gold

Tests/SwiftlyTests/HTTPClientTests.swift
kkebo: I prefer option 2 since I'd still like to verify that the Swift.org API works correctly for Ubuntu 26.04 and Debian 13.

Sources/LinuxPlatform/Linux.swift
kkebo: Fixed. d38328273365692bbfe4a1a51a1d0a77e7a5d2aa

Tests/SwiftlyTests/HTTPClientTests.swift
kkebo: Fixed.

- 2ce007712405e2c6e8d87388a9c91f10f15728af (revert commit)
- 2ad9f34765b3defb46b4ef160f3639c614b8b042 (option 2)

etcwilde (approved):
Okay, I'm happy with that. Thanks for making the changes.

PR #579, opened 2026-08-06
[1.2] Add support 6.4.x

shahmishal:


cmcgee1024:
It looks like this PR is no longer needed and can be closed since 6.4.x support was added already.

PR #588, opened 2026-08-20
1.2🍒: Init: Improve resilience on new platforms

etcwilde:
Swiftly init on platforms that don't have a release toolchain would fail and leave Swiftly in an inconsistent state. This is a quick patch to fall back on a `main-snapshot` toolchain if the latest toolchain doesn't exist.

Init will still attempt to find the latest toolchain, but if no latest toolchain exists, it will emit a warning saying so, and ask the user if they want to install a new toolchain. If they decline, swiftly will continue with the swiftly installation, but not attempt to install anything.

To make the code a little more resilient and target only cases where Swiftly failed to resolve a toolchain, I've added a new `ResolveError` type that tracks the reason for the failure and what the requested Toolchain selector was. That is used to construct the error message or to extract info about what failed.

Added tests to verify the behavior when the main snapshot install is accepted and rejected, verifying that we install the snapshot or continue to not have any installed toolchains.

(cherry picked from commit 198fc236ac18d7922451e87b3934db3533ad9b6a)
Cherry-Pick PR: https://github.com/swiftlang/swiftly/pull/587

### DIFF
--- a/Sources/Swiftly/Init.swift
+++ b/Sources/Swiftly/Init.swift
@@ -345,8 +345,35 @@ struct Init: SwiftlyCommand {
         var pathChanged = false
 
         if !skipInstall {
-            let latestVersion = try await Install.resolve(ctx, config: config, selector: ToolchainSelector.latest)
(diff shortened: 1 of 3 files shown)


etcwilde:
#587 was approved my @cmcgee1024. This change has been discussed in Webex and is desirable for Swiftly 1.2.0. Merging.

PR #592, opened 2026-09-15
Keep explicit patch version in release naming starting with Swift 6.4

shahmishal:
## Summary
- Swift.org now supports patch releases on the 6.4 branch, and its release artifacts include an explicit patch component starting there (e.g. `swift-6.4.0-RELEASE`) rather than dropping it as it did for the first release of earlier minor series (e.g. `swift-6.3-RELEASE`).
- Added `ToolchainVersion.firstReleaseWithExplicitPatch` (6.4.0) as the threshold; `ToolchainVersion.identifier` and the download URL construction in `Install.swift` now keep the patch component for any stable release `>= 6.4.0`, while releases before 6.4 keep the previous behavior unchanged.
- `swiftly install 6.4.0` now resolves/downloads/installs as `swift-6.4.0-RELEASE` instead of `swift-6.4-RELEASE`.

## Test plan
- [x] `swift build` succeeds
- [x] New `ToolchainVersionTests` verify pre-6.4 versions still drop a zero patch, and 6.4+ versions (including 6.5.0, 7.0.0) always keep it
- [x] `InstallTests` and `ToolchainSelectorTests` suites pass unchanged


Tests/SwiftlyTests/ToolchainVersionTests.swift
```diff
@@ -0,0 +1,18 @@
+import Foundation
+import SwiftlyCore
+import Testing
+
+@Suite struct ToolchainVersionTests {
+    @Test func identifierDropsPatchBeforeExplicitPatchSupport() throws {
+        #expect(ToolchainVersion(major: 5, minor: 6, patch: 0).identifier == "swift-5.6-RELEASE")
+        #expect(ToolchainVersion(major: 6, minor: 3, patch: 0).identifier == "swift-6.3-RELEASE")
+        #expect(ToolchainVersion(major: 6, minor: 3, patch: 3).identifier == "swift-6.3.3-RELEASE")
+    }
```
etcwilde, line 11: Can we add a case for `6.0.0`? If I remember correctly, that release removed the minor and patch version.

etcwilde (approved):

